PR #1, opened 2017-08-22
support passing cloud-init data as custom script

iNecas:
Currently, the cloud-init integration with ovirt supports only
a sub-set of cloud-init commands. With `custom_script` support,
one can pass arbitrary cloud-init yaml to the managed host and
make sure the provisioned host will get it unchanged (which
is not the case when passing the user data currently,
as the rbovirt parses into ovirt-specific keys and ignored the keys it
doesn't know)

Example:

```
cloudinit = <<EOF
yum_repos:
    zoo:
        baseurl: https://example.com/yumrepo
        name: example-repo
        enabled: true
EOF

vm.start_with_cloudinit(:blocking => true,
                        :user_data => cloudinit,
                        :use_custom_script => true)
```

This way, the cloudinit will get to the VM in the same format it was
passed during the call.

Requires https://github.com/abenari/rbovirt/pull/121

### DIFF
--- a/fog-ovirt.gemspec
+++ b/fog-ovirt.gemspec
@@ -22,11 +22,11 @@ Gem::Specification.new do |spec|
   spec.add_dependency("fog-core", "~> 1.45")
   spec.add_dependency("fog-json")
   spec.add_dependency("fog-xml", "~> 0.1.1")
+  spec.add_dependency("rbovirt", "~> 0.1.4")
 
   spec.add_development_dependency 'bundler', '~> 1.10'
   spec.add_development_dependency 'pry', '~> 0.10'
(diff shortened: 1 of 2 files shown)


iNecas:
the rbovrt 0.1.4 got released,with the prerequisites

adamruzicka (commented):
Not much to discuss code-wise, can't test as I don't have provisioning set up. Could we get some tests maybe?

iNecas:
Which makes me thing that having travis to run the tests against PRs would be great :)

orrabin:
Tested with latest rbovirt and this works well.

orrabin:
Thanks @iNecas!

PR #3, opened 2017-09-25
Adding operating system model

orrabin:


iNecas:
Tested and works well. Thanks @orrabin

PR #5, opened 2017-11-29
Update rubocop to 0.52

adamruzicka:
a.k.a. "touch almost every single file in the repo: the PR"

I tried to separate the fixes into commits by cop so we can easily rollback if needed.


tests/ovirt/requests/compute/create_vm_tests.rb
```diff
@@ -1,26 +1,14 @@
...
+    test("should be a kind of OVIRT::VM") { response.is_a? OVIRT::VM}
   end
 
   tests("Create VM from template (clone)") do
     response = compute.create_vm(:name => 'fog-'+(name_base+ 1).to_s, :template_name => 'hwp_small', :cluster_name => 'Default')
-    test("should be a kind of OVIRT::VM") { response.kind_of?  OVIRT::VM}
-  end
-
-  tests("Fail Creating VM") do
-    begin
```
adamruzicka, line 17: I can't wrap my head around how this was supposed to work. I'm all in for fixing it rather than deleting it, if we can figure out a way to make it test something meaningful

lib/fog/ovirt/compute.rb
```diff
@@ -79,7 +79,7 @@ module Shared
         def ovirt_attrs obj
           opts = {:raw => obj}
           obj.instance_variables.each do |v|
-            key = v.to_s.gsub("@","").to_sym
+            key = v.to_s.delete("@").to_sym
```
orrabin: Is this change related to the rubocop fix?

tests/ovirt/requests/compute/create_vm_tests.rb
orrabin: It might be a name that used to exist in the mock files we use. Maybe calling `create_vm` twice with the same parameters would actually get that error.

lib/fog/ovirt/compute.rb
adamruzicka: See http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Performance/StringReplacement

tests/ovirt/requests/compute/create_vm_tests.rb
adamruzicka: Calling it several times in a row just returns the same response

lib/fog/ovirt/compute.rb
orrabin: Thanks

orrabin:
@adamruzicka this looks good to me, could you please squash the commits?

adamruzicka:
@orrabin squashed and ready to go

.rubocop_todo.yml
```diff
@@ -0,0 +1,601 @@
+# This configuration was generated by `rubocop --auto-gen-config`
+# on 2014-05-26 15:23:01 +0100 using RuboCop version 0.22.0.
```
tbrisker: Looks like you generated the todo file (and also the main .yml) with the wrong version of fog, all the cop names have been changed.

tbrisker (changes requested):

.rubocop_todo.yml
adamruzicka: I didn't generate it at all. I copied it from `fog/fog`'s master branch to keep the styling consistent and then renamed the ones rubocop complained about

.rubocop.yml
```diff
@@ -0,0 +1,20 @@
+inherit_from: .rubocop_todo.yml
+
+HashSyntax:
+  EnforcedStyle: hash_rockets
```
tbrisker: All these rules should be changed to the update names - e.g. https://rubocop.readthedocs.io/en/latest/cops_style/#stylehashsyntax

adamruzicka:
Regenerated the `.rubocop_todo.yml` file. @tbrisker could you please take a look once more?

tbrisker (changes requested):
all the cop names in `.rubocop.yml` are old names so rubocop apparently ignores them. e.g. - `LineLength` has been replace with `Metrics/LineLength` which is now in the todo file. I suggest updating the names in the main `.rubocop.yml` and regenerating the todo.
I'm not sure how you decided which cops to fix and which not, but I think there are a lot of other easy ones in the todo.  Whether to fix everything in this PR or a separate one is another question.

orrabin:
@tbrisker could you take another look please?
the conflict is just the version bumped, it can be fixed on merge.

tbrisker (approved):
Changes make sense from cursory overview.

ShimShtein:
@adamruzicka, can you please rebase?

adamruzicka:
Rebased, both rubocop and tests should be green

ShimShtein:
@adamruzicka, looks good. I wonder why constant strings were converted from single quotes to double quotes though.
Besides that, we need to squash the commits before merging.

adamruzicka:
@ShimShtein at the beginning I copied the `.rubocop.yml` from `fog/fog`, where they have a cop enforcing double quotes enabled https://github.com/fog/fog/blob/master/.rubocop.yml#L11 . I don't really like it but this repo should probably stay consistent with `fog/fog`'s styling

ShimShtein:
Merged. Thanks @adamruzicka, now we have working rubocop!

PR #7, opened 2018-01-28
added support for OvirtError, for making things more unified for versions

ik5:


ik5:
@orrabin This is the first task we talked about, when it is accepted, I will add the supported exceptions to Foreman as well.

orrabin:
@ik5 This adds a new ovirt error but it does not replace the existing OVIRT::OvirtException.
Any error we receive from rbovirt will still get to core as an OvirtException: https://github.com/abenari/rbovirt/blob/995176066ad3d6f6d29328e8ee4f0a98a02d300e/lib/rbovirt.rb#L39

ik5:
Oh, so I'm going to make it use the new exception.

ik5:
@orrabin Now it captures all of the `OvirtException` when the `client` is called, and raising `Fog::Ovirt::Errors::OvirtError` instead.

Oh, and sorry, but there were two or three places where my OCD kicked in, and saw something like `def foo arg` that fixed to `def foo(arg)`, that's because it's something that really drives me nuts, and can't ignore, sorry for advance.

lib/fog/ovirt/models/compute/template.rb
```diff
@@ -42,11 +42,15 @@ def ready?
...
+        rescue ::OVIRT::OvirtException => e
+          raise ::Fog::Ovirt::OvertError, e.message
         end
 
         def save
-          raise Fog::Errors::Error.new('Resaving an existing object may create a duplicate') if persisted?
+          raise ::Fog::Ovirt::Errors::OvirtError, 'Providing an existing object may create a duplicate object' if persisted?
           service.client.create_template(attributes)
+        rescue ::OVIRT::OvirtException => e
+          raise ::Fog::Ovirt::OvertError, e.message
```
ShimShtein: Typo, `ovErtError` instead of `Ovirt`

lib/fog/ovirt/requests/compute/activate_volume.rb
```diff
@@ -7,6 +7,9 @@ def activate_volume(id, options)
           raise ArgumentError, "volume id is a required parameter for activate-volume" unless options.key? :id
 
           client.activate_volume(id, options[:id])
+
+        rescue ::OVIRT::OvirtException => e
+          raise ::Fog::Ovirt::OvertError, e.message
```
ShimShtein: Why are we sending only the message? I would like to see `::Fog::Ovirt::OvertError` wrapping the complete `OvirtException` instance, so I will be able to see the real source of the error.

lib/fog/ovirt/models/compute/template.rb
```diff
@@ -42,11 +42,15 @@ def ready?
...
           service.client.destroy_template(id)
+        rescue ::OVIRT::OvirtException => e
+          raise ::Fog::Ovirt::OvertError, e.message
         end
 
         def save
-          raise Fog::Errors::Error.new('Resaving an existing object may create a duplicate') if persisted?
+          raise ::Fog::Ovirt::Errors::OvirtError, 'Providing an existing object may create a duplicate object' if persisted?
           service.client.create_template(attributes)
+        rescue ::OVIRT::OvirtException => e
```
ShimShtein: Can we wrap the `client` with a generic exception handler? It would be nicer, and future client calls would be covered automatically.

I mean something like:
``` ruby
class ClientWrapper
  def initialize(client)
     @actual_client = client
  end

  def method_missing(method, *params)
    @actual_clent.send(method, *params)
    rescue ::OVIRT::OvirtException => e
      raise ::Fog::Ovirt::OvirtError, e
    end
  end
end
```

ShimShtein (changes requested):

lib/fog/ovirt/models/compute/template.rb
ik5: At the moment it does not work like that.
there is a definition of `@client` that is an ovirt sdk client, and a method that does a read only property for it named `client`.

Each place I went to, opens that specific class, and add a client.
If I understand what you are suggesting, it means to rewrite the whole concept of how it works in order to have the same way of doing stuff. But that's task number two, to my recollection.

lib/fog/ovirt/models/compute/template.rb
ik5: fixed

lib/fog/ovirt/requests/compute/activate_volume.rb
ik5: done.

lib/fog/ovirt/models/compute/template.rb
ShimShtein: My bad, this comment belongs to request class, not the model. I am talking about wrapping the [client variable](https://github.com/ik5/fog-ovirt/blob/8c311047a9e597fa29a1614196891291454b0ab6/lib/fog/ovirt/compute.rb#L143) in the request, so calls to client won't emit `::OVIRT::OvirtException` at all.

lib/fog/ovirt/models/compute/template.rb
ShimShtein: Could you add some tests to this? It looks like nobody covered that code path.

lib/fog/ovirt/models/compute/template.rb
ik5: done

ik5:
@orrabin  I'm unable to test it myself, can you give me a tip on how to it properly?

lib/fog/ovirt/core.rb
```diff
@@ -10,6 +10,7 @@ module Errors
       class ServiceError < Fog::Errors::Error; end
       class SecurityError < ServiceError; end
       class NotFound < ServiceError; end
+      class OvirtError < Fog::Errors::Error; end
```
ShimShtein, line 13: Why not splitting errors to fog-generated base and derived wrapper exception that will save the original exception object:

``` ruby
class OvirtError < Fog::Errors::Error; end

class OvirtClientError < OvirtError
  attr_reader :original
  def initialize(client_exception)
    @original = client_exception
    super("Ovirt client returned an error: #{client_exception.message}")
  end
end
```

lib/fog/ovirt/models/compute/template.rb
```diff
@@ -42,10 +42,12 @@ def ready?
 
         def destroy(options = {})
           service.client.destroy_template(id)
+        rescue ::OVIRT::OvirtException => e
```
ShimShtein: If the client is already wrapped, this should be unnecessary.

ShimShtein (changes requested):
Thanks for reformatting, a couple of comments though.

lib/fog/ovirt/models/compute/template.rb
ik5: removed it, didn't find any other locations beside `method_missing`

lib/fog/ovirt/core.rb
ik5: done

lib/fog/ovirt/compute.rb
```diff
@@ -101,6 +101,26 @@ def ovirt_attrs obj
...
       end
 
+      class ClientWrapper
+        def initialize(client)
+          @client = client
+        end
+
+        def method_missing(symbol, *args, &block)
+          if block_given?
+            yield
```
ShimShtein: Shouldn't we just pass the block to real client?

lib/fog/ovirt/compute.rb
```diff
@@ -101,6 +101,26 @@ def ovirt_attrs obj
         end
       end
 
+      class ClientWrapper
```
ShimShtein: Can you rename it to something that tells us what it is doing, `ExceptionWrapper` for example?

lib/fog/ovirt/compute.rb
```diff
@@ -140,18 +160,18 @@ def initialize(options={})
           connection_opts[:ca_no_verify]  = options[:ovirt_ca_no_verify]
           connection_opts[:filtered_api]  = options[:ovirt_filtered_api]
 
-          @client = OVIRT::Client.new(username, password, url, connection_opts)
+          @client = ClientWrapper.new(OVIRT::Client.new(username, password, url, connection_opts))
         end
 
         def api_version
           client.api_version
+        rescue ::OVIRT::OvirtException => e
```
ShimShtein: No need for this, we already do it in the `client`

lib/fog/ovirt/core.rb
```diff
@@ -10,6 +10,16 @@ module Errors
...
+
+      class OvirtClientError < OvirtError
+        attr_reader :orign_exception
```
ShimShtein: Typo? `:origN_exception` vs `@orig_exception`

ShimShtein (changes requested):
Please, pretty please, can you add `shindo` tests for all this new functionality?

lib/fog/ovirt/core.rb
ik5: good catch

lib/fog/ovirt/compute.rb
ik5: opps, missed that point, sorry

lib/fog/ovirt/compute.rb
ik5: not sure, because the block/proc part will be belonging to the wrapper, for something to do. Not sure if it should be sent to the real client.

lib/fog/ovirt/compute.rb
ik5: done

ik5:
@ShimShtein, still trying to figure out how to test it.
The Mock part ignores the client completely, and trying to see how to make a mock that does use it, so I can test it.

lib/fog/ovirt/compute.rb
ShimShtein: The binding of the block is to the caller, but I think that it is intended to run in the context of the client anyway. It can really be a problem, consider:

``` ruby
# rbovirt client implementation:
class Client
  def inner_throw_method
    throw ::OVIRT::OvirtException
  end

  def outer_method
    puts 'this should run first'
    yield
  end
end

# Fog request implementation
class FogOvirtRequest
  def do_something
    client.outer_method do
      inner_throw_method
    rescue ::OVIRT::OvirtException
      log 'This is expected' 
    end
  end
end
```

It will fail with the your suggested implementation for two reasons: the exception would be wrapped and the order will be changed.

ik5: i see your point there, you are right, changing it

ik5:
@ShimShtein here are the tests :)

tests/ovirt/requests/compute/client_tests.rb
```diff
@@ -0,0 +1,16 @@
...
+    raise OVIRT::OvirtException, 'Test'
+  end
+
+  begin
+    object_under_test.foo
+  rescue Fog::Ovirt::Errors::OvirtClientError => e
+    error = e
+  end
+
+  tests('Has valid backtrace').returns(error.nil?) { false }
```
ShimShtein: Almost there :smile: 
I think we can skip the backtrace test altogether, instead, can you please test that:

1. Exception properly wraps the text: `error.message == 'Ovirt client returned an error: '`
2. Exception includes the original erorr: `error.orig_exception.is_a?(OVIRT::OvirtException) && error.orig_exception.message == 'Test'`

ik5: The second test exists, but forgot to commit it (sorry).

I think that your suggestion for the first test is actually much better then what I was thinking in doing :)

ik5:
following some shindo examples, found a nicer way to do it :)

tests/ovirt/requests/compute/client_tests.rb
```diff
@@ -0,0 +1,23 @@
...
+    @object_under_test.foo
+  end
+
+  tests('The exception test is as expected').returns(true) do
+    begin
+      @object_under_test.foo
+    rescue Fog::Ovirt::Errors::OvirtClientError => e
+      e.message == 'Ovirt client returned an error: Test'
+    end
+  end
```
ShimShtein: Can you also test that you get the original exception in `orig_exception` (2 in my previous comment)?

ik5: sorry, my bad, here it is :)

ShimShtein (approved):
:+1:  ACK from me.

orrabin:
@ik5 please squash the commits

lib/fog/ovirt/version.rb
```diff
@@ -1,5 +1,5 @@
 module Fog
   module Ovirt
-    VERSION = '0.1.2'
+    VERSION = '0.1.3'.freeze
```
orrabin: why the version bump? why freeze?

orrabin (commented):
Tested by:
- debugging the `test_connection` method does get the correct message with this pr:
``` #<Fog::Ovirt::Errors::OvirtClientError: Ovirt client returned an error: 401 Unauthorized>```
- trying to save a vm with no cores or memory returns ` Fog::Ovirt::Errors::OvirtClientError`

File: lib/fog/ovirt/version.rb (hunk 12 above)
Comment (ik5) [reply lib/fog/ovirt/version.rb]: > why the version bump? 

It changes how exception works, so I think it should be a new version, so exception handlers will need to be changed.

> why freeze?

Ruby's constant are mutable by defaults. Moreover, each time you call a constant string, it allocate new instance of that string.
Calling `freeze` does two main things here:
1. Makes the constant immutable (more or less)
2. Use the constant allocation once, rather then allocate new memory each usage.

**Comment by ik5** [comment]
I'ved squashed everything, and no version bump

File: lib/fog/ovirt/version.rb [hunk 13]
```diff
@@ -1,5 +1,5 @@
 module Fog
   module Ovirt
-    VERSION = '0.1.2'
+    VERSION = '0.1.2'.freeze
```
orrabin: Please move this to a separate pr, there are now merge conflicts because the version was bumped before this pr was merged.

ik5: sorry on that :'( removed the freeze

orrabin: The freeze was a good change, just didn't want it to affect the rest of the pr.
It's already suggested in: https://github.com/fog/fog-ovirt/pull/5/files#diff-73b69b006fbdd314071398980a75b3eaR3

lib/fog/ovirt/core.rb
```diff
@@ -10,6 +10,16 @@ module Errors
       class ServiceError < Fog::Errors::Error; end
       class SecurityError < ServiceError; end
       class NotFound < ServiceError; end
+      class OvirtError < Fog::Errors::Error; end
+
+      class OvirtClientError < OvirtError
```
orrabin: This is a little confusing, `OvirtClient` does not show this is an actual error we received from Ovirt.
Maybe this can be renamed to `OvirtEngineError`?
My first impression was that `OvirtError` came from ovirt so I didn't understand it was the base class.

ik5: here is a rename, tested it as well.

orrabin (approved):
LGTM

ShimShtein:
Merged, thanks @ik5!

PR #10, opened 2018-03-20
Adding support for api v4

orrabin:
Not all requests have been translated yet. Tests are passing but there are still some fixes to get the full workflow to work.
Opening to allow some early reviews and testing.


lib/fog/ovirt/requests/compute/v4/add_volume.rb
```diff
@@ -0,0 +1,54 @@
+module Fog
+  module Compute
+    class Ovirt
+      class V4
+        class Real
+          # rubocop:disable Style/NumericLiterals
+          DISK_SIZE_TO_GB = 1073741824
```
iNecas: Writing this as `DISK_SIZE_TO_GB = 1024 * 1024 * 1024` would be a bit more readable

lib/fog/ovirt/requests/compute/v4/add_volume.rb
```diff
@@ -0,0 +1,54 @@
...
+          # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+          def add_options_defaults(options)
+            search = options[:search] || format("datacenter=%<datacenter>s", :datacenter => datacenter)
+            options[:bootable] = "true" unless options[:bootable]
+            options[:interface] = OvirtSDK4::DiskInterface::VIRTIO unless options[:interface].present?
+            options[:provisioned_size] = options[:size_gb].to_i * DISK_SIZE_TO_GB if options[:size_gb]
+
+            options[:storage_domain_id] = options[:storage_domain] || storagedomains(:role => "data", :search => search).first.id
+            # If no size is given, default to a volume size of 8GB
+            options[:provisioned_size] = 8 * 1024 * 1024 * 1024 unless options[:provisioned_size]
```
iNecas: We should probably use `DISK_SIZE_TO_GB` here as well

orrabin:
This is now ready for review

orrabin:
@iNecas I fixed the destroy method.

lib/fog/ovirt/compute/v4.rb
```diff
@@ -0,0 +1,161 @@
...
+
+            connection_opts = {
+              :url      => url,
+              :username => username,
+              :password => password
+            }
+
+            @datacenter = options[:ovirt_datacenter]
+            connection_opts[:ca_file]  = options[:ca_file]
+            connection_opts[:ca_certs] = [OpenSSL::X509::Certificate.new(options.delete(:public_key))] if options[:public_key].present?
```
iNecas: I guess we should not modify the `options` hash and use `options[:public_key]` instead of `options.delete(:public_key)`

iNecas:
I would like to do some more testing, but in general close to get in. One thing, that could be done either inside this PR or as part of another one, would be some table describing the status of v4 support in terms of supported models/requests.

lib/fog/ovirt/compute.rb
```diff
@@ -48,15 +50,23 @@ def respond_to?(symbol, include_all = false)
       end
 
       require "fog/ovirt/compute/v3"
+      require "fog/ovirt/compute/v4"
```
ShimShtein, line 54: We probably want to move the requires to `self.new` initializer. We don't want to load V3 API if the user has requested V4. I would suggest moving it inside `if options[:api_version]` block:

```ruby
        if options[:api_version] == "v4"
          require "fog/ovirt/compute/v4"
          @client = Fog::Compute::Ovirt::V4.new(options)
        else
          require "fog/ovirt/compute/v3"
          @client = Fog::Compute::Ovirt::V3.new(options)
        end
```

lib/fog/ovirt/compute.rb
```diff
@@ -48,15 +50,23 @@ def respond_to?(symbol, include_all = false)
...
 
       def self.new(options = {})
-        super()
-        @client = Fog::Compute::Ovirt::V3.new(options)
+        super(options)
```
ShimShtein, line 57: If we are passing options to the base class (`Fog::Service`), we should sanitize options that are not relevant to it:

```ruby
api_version = options.delete(:api_version)
if api_version == "v4"
  sanitized_options = Fog::Compute::Ovirt::V4.sanitize_options(options)
end

super(sanitized_options)
```

lib/fog/ovirt/compute/v4.rb
```diff
@@ -0,0 +1,161 @@
+module Fog
+  module Compute
+    class Ovirt
+      class V4 < Fog::Service
+        requires   :ovirt_username, :ovirt_password
+        recognizes :ovirt_url,      :ovirt_server, :ovirt_port, :ovirt_api_path, :ovirt_datacenter,
+                   :ca_file, :public_key, :api_version
```
ShimShtein: Why does it need to recognize `:api_version`?

lib/fog/ovirt/compute/v4.rb
```diff
@@ -0,0 +1,161 @@
...
+
+          # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity
+          def get_attr_value(value, opts)
+            case value
+            when OvirtSDK4::TemplateVersion, Array, Hash, OvirtSDK4::List, OvirtSDK4::DataCenter
+              value
+            when OvirtSDK4::Mac
+              value.address
+            when OvirtSDK4::Cpu
+              opts[:cores] = value.topology.nil? ? nil : value.topology.cores
```
ShimShtein, line 82: Why are we setting a property directly in `opts` from here? It should be a function - no changes to inputs should be made. This will remove the need to `opts` parameter altogether. I would expect:

```ruby
value.topology.try(:cores)
```

If we don't know what would be the key of the property, we should add it as a special case to `ovirt_attrs`:

``` ruby
            obj.instance_variables.each do |v|
              key = v.to_s.delete("@").to_sym
              value = obj.instance_variable_get(v)
              
              opts[:cores] = get_attr_value(value) if value.is_a? OvirtSDK4::Cpu
```

lib/fog/ovirt/compute/v4.rb
```diff
@@ -0,0 +1,161 @@
...
+
+          # rubocop:disable Metrics/AbcSize
+          def initialize(options = {})
+            require "ovirtsdk4"
+            username   = options[:ovirt_username]
+            password   = options[:ovirt_password]
+            server     = options[:ovirt_server]
+            port       = options[:ovirt_port]       || 8080
+            api_path   = options[:ovirt_api_path]   || "/api"
+            url        = options[:ovirt_url]        || "#{@scheme}://#{server}:#{port}#{api_path}"
```
ShimShtein: I don't see where we define `@scheme`, I think it will always default to `nil`.

lib/fog/ovirt/requests/compute/v4/add_volume.rb
```diff
@@ -0,0 +1,53 @@
...
+
+            disk_attachments_service = client.system_service.vms_service.vm_service(id).disk_attachments_service
+            disk = OvirtSDK4::DiskAttachment.new(options)
+            disk_attachments_service.add(disk)
+          end
+
+          # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+          def add_options_defaults(options)
+            search = options[:search] || format("datacenter=%<datacenter>s", :datacenter => datacenter)
+            options[:bootable] = "true" unless options[:bootable]
```
ShimShtein:
``` ruby
options[:bootable] ||= "true"
```

lib/fog/ovirt/requests/compute/v4/add_volume.rb
```diff
@@ -0,0 +1,53 @@
...
+            options[:interface] = OvirtSDK4::DiskInterface::VIRTIO unless options[:interface].present?
```
ShimShtein:
``` ruby
options[:interface] ||= OvirtSDK4::DiskInterface::VIRTIO
```

The same for all other properties are setting default, if the property is not present.

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -0,0 +1,51 @@
+module Fog
+  module Compute
+    class Ovirt
+      class V4
+        class Real
+          # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+          def process_vm_opts(opts)
+            return unless (opts[:template] || opts[:template_name]) && (opts[:storagedomain] || opts[:storagedomain_name])
```
ShimShtein: please refactor to a method:

``` ruby
return unless has_option?(opts, 'template') && has_option?(opts, 'storagedomain')

def has_option?(opts, name)
  opts[name.to_sym] || opts[(name + '_name').to_sym]
end
```

lib/fog/ovirt/requests/compute/v4/storage_domains.rb
```diff
@@ -0,0 +1,25 @@
+module Fog
+  module Compute
+    class Ovirt
+      class V4
+        class Real
+          def storage_domains(filter = {})
+            client.system_service.storage_domains_service.list.collect do |sd|
```
ShimShtein: Except for roles, there are other search options, remove roles from filter, and pass the filter to `list`. After that you can deal with roles in code.

lib/fog/ovirt/requests/compute/v4/update_interface.rb
```diff
@@ -0,0 +1,37 @@
...
+          end
+        end
+
+        class Real
+          extend ::Fog::Compute::Ovirt::V3::Shared
+
+          def update_interface(id, options)
+            check_arguments(id, options)
+
+            interface_id = options[:id]
```
ShimShtein, line 18: it can be a single line: 
``` ruby
interface_id = options.delete(:id)
```

lib/fog/ovirt/requests/compute/v4/update_volume.rb
```diff
@@ -0,0 +1,38 @@
...
+          end
+        end
+
+        class Real
+          extend ::Fog::Compute::Ovirt::V3::Shared
+
+          def update_volume(id, options)
+            check_arguments(id, options)
+
+            disk_id = options[:id]
```
ShimShtein, line 18: It can be a single line

ShimShtein (changes requested):

lib/fog/ovirt/requests/compute/v4/vm_action.rb
```diff
@@ -0,0 +1,26 @@
...
+  module Compute
+    class Ovirt
+      class V4
+        class Real
+          def vm_action(options = {})
+            raise ArgumentError, "instance id is a required parameter" unless options.key? :id
+            raise ArgumentError, "action is a required parameter" unless options.key? :action
+
+            vm_service = client.system_service.vms_service.vm_service(options[:id])
+            vm_service.public_send(options[:action])
```
ShimShtein, line 11: @iNecas, this looks like a security issue to me - the parameter is not sanitized and sent directly to the object. I know this is not intended to be called directly by a user, but I think it deserves at least a warning comment in the code. I think the developer should be warned that the input is not sanitized inside, so it's his responsibility to sanitize the action before sending it to here.

lib/fog/ovirt/compute/v4.rb
orrabin: because it's not deleted from the initialization options

lib/fog/ovirt/compute/v4.rb
orrabin: Because we have to do this to keep the model like the v3 one.
The `try` is not defined for topology, I tried that.

```opts[:cores] = get_attr_value(value) if value.is_a? OvirtSDK4::Cpu```
 would not remove the ` when OvirtSDK4::Cpu` in `get_attr_value` as it still needs to get `value.topology.cores`

**Comment by orrabin** [comment]
> One thing, that could be done either inside this PR or as part of another one, would be some table describing the status of v4 support in terms of supported models/requests.

I agree but where should it be? maybe a section in the readme?

File: lib/fog/ovirt/requests/compute/v4/update_interface.rb (hunk 13 above)
Comment (orrabin) [reply lib/fog/ovirt/requests/compute/v4/update_interface.rb:18]: leftovers from testing, it can actually be on line 21

File: lib/fog/ovirt/requests/compute/v4/add_volume.rb (hunk 9 above)
Comment (orrabin) [reply lib/fog/ovirt/requests/compute/v4/add_volume.rb]: This is all copied from rbovirt, I'll fix it.

File: lib/fog/ovirt/compute.rb (hunk 4 above)
Comment (orrabin) [reply lib/fog/ovirt/compute.rb:54]: I don't remember why right now but if I don't require these before the initialize if fails initialization.

File: lib/fog/ovirt/requests/compute/v4/vm_action.rb (hunk 15 above)
Comment (orrabin) [reply lib/fog/ovirt/requests/compute/v4/vm_action.rb:11]: I don't think it's such a security issue because it's only public methods and we allow other actions like `destory_vm`.
We could change this to only allow power actions since that's why foreman core uses it but there might be other actions acceptable here/

**Comment by orrabin** [comment]
@ShimShtein thanks for the review!
I fixed most of the comments.
There are a few that need more testing so they might be moved to a refactoring issue to allow merging this as a first experimental version.

**Comment by iNecas** [comment]
Section in README would be just fine

File: lib/fog/ovirt/requests/compute/v4/vm_action.rb (hunk 15 above)
Comment (iNecas) [reply lib/fog/ovirt/requests/compute/v4/vm_action.rb:11]: Just a comment in the documentation of the method to warn about this behavior should be enough there.  I agree with @orrabin that we might not be sure on what all all the methods that should actually be allowed/denied.

File: lib/fog/ovirt/compute.rb (hunk 5 above)
Comment (iNecas) [reply lib/fog/ovirt/compute.rb:57]: Let's not suggest using `delete` against options passed as arguments, as they can cause surprises for the user developer, finding that the hash that was passed as argument to some method was suddenly changed.

File: lib/fog/ovirt/requests/compute/v4/add_volume.rb [hunk 16]
```diff
@@ -0,0 +1,53 @@
...
+            options[:bootable] ||= "true"
+            options[:interface] ||= OvirtSDK4::DiskInterface::VIRTIO
+            options[:provisioned_size] = options[:size_gb].to_i * DISK_SIZE_TO_GB if options[:size_gb]
+
+            options[:storage_domain_id] = options[:storage_domain] || storagedomains(:role => "data", :search => search).first.id
+            # If no size is given, default to a volume size of 8GB
+            options[:provisioned_size] ||= 8 * DISK_SIZE_TO_GB
+            options[:type] ||= OvirtSDK4::DiskType::DATA
+            options[:format] ||= OvirtSDK4::DiskFormat::COW
+            options[:sparse] ||= true
```
iNecas: This way, one could never pass `:sparse => false`, as this line would always overwrite it with `true`

iNecas: We need to change that back to something like `options[:sparse] = true unless options.key?(:sparse)`. Alternatively, we could do something like

```ruby
options = options.reverse_merge(:sparse => true, :format => OvirtSDK4::DiskFormat::COW…)
```

iNecas: Scratch the `reverse_merge`: it's in rails, not in pure ruby

lib/fog/ovirt/requests/compute/v4/add_volume.rb
```diff
@@ -0,0 +1,53 @@
...
+            options[:quota] = options[:quota].present? ? client.system_service.data_centers_service.data_center_service(datacenter).quotas_service.quota_service(options[:quota]).get : nil
+
+            options[:disk] ||= {}
+            options[:disk][:storage_domains] ||= [client.system_service.storage_domains_service.storage_domain_service(options[:storage_domain_id]).get]
+            options[:disk][:provisioned_size] ||= options.delete(:provisioned_size)
+            options[:disk][:type] ||= options.delete(:type)
+            options[:disk][:format] ||= options.delete(:format)
+            options[:disk][:sparse] ||= options.delete(:sparse)
```
iNecas: Same as with `options[:sparse]`, this will not work, as we work with booleans

lib/fog/ovirt/requests/compute/v4/add_volume.rb
```diff
@@ -0,0 +1,53 @@
...
+            options = add_options_defaults(options)
+
+            disk_attachments_service = client.system_service.vms_service.vm_service(id).disk_attachments_service
+            disk = OvirtSDK4::DiskAttachment.new(options)
+            disk_attachments_service.add(disk)
+          end
+
+          # rubocop:disable Metrics/AbcSize
+          def add_options_defaults(options)
+            search = options[:search] || format("datacenter=%<datacenter>s", :datacenter => datacenter)
```
iNecas: Since we are modifying the options argument, we should do `options = options.dup` at the beginning of the method

lib/fog/ovirt/requests/compute/v4/list_virtual_machines.rb
```diff
@@ -0,0 +1,26 @@
+module Fog
+  module Compute
+    class Ovirt
+      class V4
+        class Real
+          def list_virtual_machines(filters = {})
+            page = filters.delete(:page)
```
iNecas: Ditto regarding `.delete` method

lib/fog/ovirt/requests/compute/v4/storage_domains.rb
```diff
@@ -0,0 +1,26 @@
+module Fog
+  module Compute
+    class Ovirt
+      class V4
+        class Real
+          def storage_domains(filter = {})
+            role_filter = filter.delete(:role)
```
iNecas: Ditto regarding `.delete` method

lib/fog/ovirt/requests/compute/v4/update_interface.rb
```diff
@@ -0,0 +1,35 @@
...
+          end
+        end
+
+        class Real
+          extend ::Fog::Compute::Ovirt::V4::Shared
+
+          def update_interface(id, options)
+            check_arguments(id, options)
+
+            interface_id = options.delete(:id)
```
iNecas: Ditto regarding `.delete` method

lib/fog/ovirt/requests/compute/v4/update_volume.rb
```diff
@@ -0,0 +1,36 @@
...
+          end
+        end
+
+        class Real
+          extend ::Fog::Compute::Ovirt::V4::Shared
+
+          def update_volume(id, options)
+            check_arguments(id, options)
+
+            disk_id = options.delete(:id)
```
iNecas: Ditto regarding `.delete` method

iNecas: If we don't want to pass the `:id` later, we should do `options = options.dup` at the beginning`

lib/fog/ovirt/requests/compute/v4/add_interface.rb
```diff
@@ -0,0 +1,40 @@
...
+            if options[:network].present?
+              network = client.system_service.networks_service.network_service(options[:network]).get
+
+              profiles = client.follow_link(network.vnic_profiles)
+
+              profile = profiles.detect { |x| x.name == network.name }
+
+              profile ||= profiles.sort_by(&:name).first
+
+              options.delete(:network)
```
iNecas, line 22: Ditto regarding `.delete` method

lib/fog/ovirt/requests/compute/v4/storage_domains.rb
orrabin: This one is on purpose, see comment below

orrabin: I mean I have to delete it or I can't send the filter to the list

lib/fog/ovirt/requests/compute/v4/add_interface.rb
orrabin: I added options = options.dup on line 11

lib/fog/ovirt/requests/compute/v4/add_volume.rb
```diff
@@ -0,0 +1,54 @@
...
+            options[:format] ||= OvirtSDK4::DiskFormat::COW
+            options[:sparse] = true unless options[:sparse].present?
+            options[:quota] = options[:quota].present? ? client.system_service.data_centers_service.data_center_service(datacenter).quotas_service.quota_service(options[:quota]).get : nil
+
+            options[:disk] ||= {}
+            options[:disk][:storage_domains] ||= [client.system_service.storage_domains_service.storage_domain_service(options[:storage_domain_id]).get]
+            options[:disk][:provisioned_size] ||= options.delete(:provisioned_size)
+            options[:disk][:type] ||= options.delete(:type)
+            options[:disk][:format] ||= options.delete(:format)
+            options[:disk][:sparse] = options.delete(:sparse) unless options[:sparse].present?
```
iNecas: Shouldn't this be:

```ruby
options[:disk][:sparse] = options.delete(:sparse) unless options[:disk][:sparse].present?
```

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -0,0 +1,55 @@
...
+          end
+          # rubocop:enable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+
+          # rubocop:disable Metrics/AbcSize
+          def create_vm(attrs)
+            attrs[:cluster_name] ||= datacenter.clusters.first.name unless attrs[:cluster]
+            vms_service = client.system_service.vms_service
+            attrs[:cluster] = client.system_service.clusters_service.cluster_service(attrs[:cluster]).get
+            attrs[:instance_type] = attrs[:instance_type].present? ? client.system_service.instance_types_service.instance_type_service(attrs[:instance_type]).get : nil
+            attrs[:template] = attrs[:template].present? ? client.system_service.templates_service.template_service(attrs[:template]).get : nil
```
iNecas: The `nil` is not a valid value for blank template, what was working for me was:

```ruby
            attrs[:template] = if attrs[:template].present?
                                 client.system_service.templates_service.template_service(attrs[:template]).get
                               else
                                 client.system_service.get.special_objects.blank_template
                               end
```

orrabin:
@iNecas I fixed the templates.

Known issues when testing with core: quota is only filled after saving CR (or if data center is changed after loading data centers), the issue you had with networks.

I'll add a separate PR to the readme file regarding which requests for v4 are now supported and that it is a first version so issues are welcome.

@ShimShtein I did not finish all the refactoring you requested, we should open an issue for them.

iNecas (approved):

iNecas:
Agreed. Tested and was able to provision the host using the v4. We can address the rough edges as we go, but there is no point in holding this any longer. Thanks @orrabin. Could you release a new version? I will then bump the dependency in https://github.com/theforeman/foreman/pull/5382

iNecas:
I realized that I actually have permissions to push new version: I've released 1.0.2

PR #15, opened 2018-04-09
Fix core number sent in v4

orrabin:
Cores was not sent correctly so ovirt always defaulted to 1 core.
This fixes the cores to be sent as part of the Cpu details.

### DIFF
--- a/lib/fog/ovirt/requests/compute/v4/create_vm.rb
+++ b/lib/fog/ovirt/requests/compute/v4/create_vm.rb
@@ -41,6 +41,11 @@ module Fog
             attrs[:comment] ||= ""
             attrs[:quota] = attrs[:quota].present? ? client.system_service.data_centers_service.data_center_service(datacenter).quotas_service.quota_service(attrs[:quota]).get : nil
 
+            if attrs[:cores].present?
+              cpu_topology = OvirtSDK4::CpuTopology.new(:cores => attrs[:cores], :sockets => "1")
+              attrs[:cpu] = OvirtSDK4::Cpu.new(:topology => cpu_topology)
+            end
+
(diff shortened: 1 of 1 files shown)


shiramax:
I tested it, and it look OK to me :+1:

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -41,6 +41,9 @@ def create_vm(attrs)
             attrs[:comment] ||= ""
             attrs[:quota] = attrs[:quota].present? ? client.system_service.data_centers_service.data_center_service(datacenter).quotas_service.quota_service(attrs[:quota]).get : nil
 
+            cpu_topology = OvirtSDK4::CpuTopology.new(:cores => attrs[:cores], :sockets => "1") if attrs[:cores].present?
+            attrs[:cpu] = OvirtSDK4::Cpu.new(:topology => cpu_topology)
```
adamruzicka: I'd be slightly inclined towards not assigning anything to `attrs[:cpu]` in case `attrs[:cores]` is `#blank?` and letting `OvirtSDK4` fill in the defaults, but from what I've seen in OSDK4 it shouldn't really matter.

+0 for having this done conditionally

adamruzicka (approved):
Left one nitpick inline, otherwise looks good to me

lib/fog/ovirt/requests/compute/v4/create_vm.rb
orrabin: I'll fix it now

iNecas:
Thanks @orrabin, @shiramax and @adamruzicka

PR #16, opened 2018-04-10
 Add Socket to server

shiramax:
This pr follows https://github.com/fog/fog-ovirt/pull/15,
I'll rebase once it will be merged.

### DIFF
--- a/lib/fog/ovirt/compute/v4.rb
+++ b/lib/fog/ovirt/compute/v4.rb
@@ -80,6 +80,7 @@ module Fog
               value.address
             when OvirtSDK4::Cpu
               opts[:cores] = value.topology.nil? ? nil : value.topology.cores
+              opts[:sockets] = value.topology.nil? ? nil : value.topology.sockets
             when OvirtSDK4::Display
               subject = value.certificate.subject if value.certificate
               {
--- a/lib/fog/ovirt/requests/compute/v4/create_vm.rb
+++ b/lib/fog/ovirt/requests/compute/v4/create_vm.rb
@@ -40,9 +40,8 @@ module Fog
 
             attrs[:comment] ||= ""
             attrs[:quota] = attrs[:quota].present? ? client.system_service.data_centers_service.data_center_service(datacenter).quotas_service.quota_service(attrs[:quota]).get : nil
-
-            if attrs[:cores].present?
-              cpu_topology = OvirtSDK4::CpuTopology.new(:cores => attrs[:cores], :sockets => "1")
+            if attrs[:cores].present? || attrs[:sockets].present?
+              cpu_topology = OvirtSDK4::CpuTopology.new(:cores => attrs.fetch(:cores, "1"), :sockets => attrs.fetch(:sockets, "1"))
               attrs[:cpu] = OvirtSDK4::Cpu.new(:topology => cpu_topology)
             end
 
--- a/lib/fog/ovirt/models/compute/server.rb
+++ b/lib/fog/ovirt/models/compute/server.rb
@@ -19,6 +19,7 @@ module Fog
         attribute :ip
         attribute :status
         attribute :cores, :aliases => "cpus"
+        attribute :sockets
         attribute :memory
         attribute :host
         attribute :cluster


shiramax:
This will only work in API V3 once https://github.com/abenari/rbovirt/pull/123 is merged, and a new version of rbovirt is released

shiramax:
Looks OK to me :+1:

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -40,14 +40,21 @@ def create_vm(attrs)
 
             attrs[:comment] ||= ""
             attrs[:quota] = attrs[:quota].present? ? client.system_service.data_centers_service.data_center_service(datacenter).quotas_service.quota_service(attrs[:quota]).get : nil
+            cpu_topology = OvirtSDK4::CpuTopology.new(:cores => attrs[:cores], :sockets => attrs[:sockets])
+
+            if attrs[:cores].present? || attrs[:sockets].present?
+              attrs[:cpu] = OvirtSDK4::Cpu.new(:topology => cpu_topology)
+            end
+
+            cpu_topology = OvirtSDK4::CpuTopology.new(:cores => attrs[:cores], :sockets => "1") if attrs[:cores].present?
```
ekohl: Here you are ignoring the sockets. Is that intentional? If so, I'd create one if/else block and assign `attrs[:cpu]` just once. The rubocop is IMHO right and it's hard to exactly follow what's going.

shiramax: @ekohl , you are right, this line shouldn't be here.

shiramax:
@ekohl , please review again :)

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -40,14 +40,18 @@ def create_vm(attrs)
 
             attrs[:comment] ||= ""
             attrs[:quota] = attrs[:quota].present? ? client.system_service.data_centers_service.data_center_service(datacenter).quotas_service.quota_service(attrs[:quota]).get : nil
+            cpu_topology = OvirtSDK4::CpuTopology.new(:cores => attrs[:cores], :sockets => attrs[:sockets])
```
ekohl: You can move this into the if-statement below since it's not needed otherwise.

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -24,7 +24,7 @@ def process_vm_opts(opts)
           end
           # rubocop:enable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
 
-          # rubocop:disable Metrics/AbcSize
+          # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
```
ekohl: With this simpler code, is it still needed to disable these cops?

lib/fog/ovirt/requests/compute/v4/create_vm.rb
shiramax: DONE

lib/fog/ovirt/requests/compute/v4/create_vm.rb
shiramax: yes, rubocop still wants to disable it.

ekohl (approved):
I think this makes sense, but I'm not that familiar with this codebase so I think it needs another reviewer :)

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -41,8 +41,8 @@ def create_vm(attrs)
             attrs[:comment] ||= ""
             attrs[:quota] = attrs[:quota].present? ? client.system_service.data_centers_service.data_center_service(datacenter).quotas_service.quota_service(attrs[:quota]).get : nil
 
-            if attrs[:cores].present?
-              cpu_topology = OvirtSDK4::CpuTopology.new(:cores => attrs[:cores], :sockets => "1")
+            if attrs[:cores].present? || attrs[:sockets].present?
+              cpu_topology = OvirtSDK4::CpuTopology.new(:cores => attrs[:cores], :sockets => attrs[:sockets])
```
orrabin: There should probably be a default here in case one of them exists and the other is nil, maybe:
` :cores => attrs[:cores] || "1" , :sockets => attrs[:sockets] || "1"`

adamruzicka: Just curious, what is the semantic between cores and sockets? My assumption would be that a machine has to have at least as many cores as it has sockets. Is it correct and does it still hold in ovirt?

If that is true, we probably shouldn't default to 1 for cores

ekohl: I vaguely recall it was cores per socket so 2 cores and 2 sockets means 4 total CPUs. https://www.ovirt.org/develop/release-management/features/virt/hot-plug-cpu/#ui does imply that. In that case 1*1 sounds like a sane default.

shiramax:
@orrabin , 
Thanks, please review again :)

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -40,9 +40,8 @@ def create_vm(attrs)
 
             attrs[:comment] ||= ""
             attrs[:quota] = attrs[:quota].present? ? client.system_service.data_centers_service.data_center_service(datacenter).quotas_service.quota_service(attrs[:quota]).get : nil
-
-            if attrs[:cores].present?
-              cpu_topology = OvirtSDK4::CpuTopology.new(:cores => attrs[:cores], :sockets => "1")
+            if attrs[:cores].present? || attrs[:sockets].present?
+              cpu_topology = OvirtSDK4::CpuTopology.new(:cores => attrs[:cores] || "1", :sockets => attrs[:sockets] || "1")
```
ShimShtein: What will happen if I have defined 2 sockets and no cores?

ShimShtein: Also you can use [#fetch](https://ruby-doc.org/core-2.2.0/Hash.html#method-i-fetch) method to simplify the code:

``` ruby
cpu_topology = OvirtSDK4::CpuTopology.new(:cores => attrs.fetch(:cores, "1"), :sockets => attrs.fetch(:sockets, "1"))
```

lib/fog/ovirt/compute/v4.rb
```diff
@@ -80,6 +80,7 @@ def get_attr_value(value, opts)
               value.address
             when OvirtSDK4::Cpu
               opts[:cores] = value.topology.nil? ? nil : value.topology.cores
+              opts[:sockets] = value.topology.nil? ? nil : value.topology.sockets
```
ShimShtein, line 83: We can use `#try` here:

``` ruby
opts[:cores] = value.topology.try(:cores)
opts[:sockets] = value.topology.try(:sockets)
```

ShimShtein (changes requested):

lib/fog/ovirt/compute/v4.rb
shiramax: try is rails function, it can't be used in fog-ovirt.

lib/fog/ovirt/requests/compute/v4/create_vm.rb
shiramax: if there are 2 sockets and no cores then the core amount should be 1.

iNecas:
@ShimShtein seems like needs re-review

ShimShtein (approved):

ShimShtein:
Merged.
Thanks @shiramax!

PR #17, opened 2018-04-10
Fixes #23194 - Add support to VM with memory smaller then 1 GB

shiramax:


orrabin:
Works well for me, thank you for also refactoring the memory calculation :)

orrabin:
merged, thanks @shiramax !

Issue #19, opened 2018-04-25
NameError: uninitialized constant Fog::JSON


lhellebr:
When using Foreman with Fog and Fog-oVirt (Create Host -> select oVirt -> OS -> Image Based), I get the attached traceback.
[tb.txt](https://github.com/fog/fog-ovirt/files/1947059/tb.txt)

lhellebr:
Most likely, JSON should be required in both fog-ovirt and fog-core.

lhellebr:
Related: https://github.com/fog/fog/issues/3993 , https://github.com/fog/fog-core/issues/234

geemus:
I think this will likely be fixed by #20

lhellebr:
Although the PR ^ is still not merged, I was unable to reproduce this today.

shiramax:
The issue was fixed by: https://github.com/fog/fog-ovirt/pull/20

geemus:
Sounds good, thanks!

PR #20, opened 2018-04-26
Add require fog-json

### LINKED ISSUE #234: NameError: uninitialized constant Fog::JSON
When using Foreman with Fog and Fog-oVirt (Create Host -> select oVirt -> OS -> Image Based), I get the attached traceback.
[tb.txt](https://github.com/fog/fog-core/files/1947120/tb.txt)

Related: https://github.com/fog/fog-ovirt/issues/19 , https://github.com/fog/fog/issues/3993

orrabin:
It makes more sense to actually fix: https://github.com/fog/fog-core/issues/234 and add a require when using Fog::JSON but for now this workaround works.

### DIFF
--- a/lib/fog/ovirt.rb
+++ b/lib/fog/ovirt.rb
@@ -1,5 +1,6 @@
 require "fog/core"
 require "fog/xml"
+require "fog/json"
 
 module Fog
   module Compute


iNecas:
Rubocop is not happy, ACK besides that.

orrabin:
empty newline was missing :) updated

lhellebr:
Although this PR is still not merged, I was not able to reproduce https://github.com/fog/fog-ovirt/issues/19 today.

iNecas:
Rubocop is still not happy, perhaps rebase needed?

ShimShtein:
I can't find where do we use functionality from `fog-json`.
@lhellebr, were you able to reproduce it?

lhellebr:
See the original bug report for traceback (so yes, I was able to reproduce it). I wasn't, however, able to reproduce it downstream some time later so something else might have changed in the meantime.

iNecas:
We require that via some fog-core dependency, that we use on the backend. I can confirm that this change actually fixed the issue for me.

orrabin:
@iNecas I fixed the rubocop errors

lib/fog/ovirt.rb
```diff
@@ -1,5 +1,6 @@
 require "fog/core"
 require "fog/xml"
+require "fog-json"
```
iNecas: the tests are failing currently: shouldn't this be 'fog/json'?

orrabin: yeah it should, it got messed up when resolving merge conflicts

adamruzicka:
This is good to go, isn't it?

iNecas:
It would be good to rebase, just to see if the tests are still passing, just in case

orrabin:
@iNecas, @adamruzicka I rebased the pr.

iNecas:
Seems like rubocop needs updating.

orrabin:
@iNecas it's not related to this PR, it's new warnings from rubocop 0.59.1.
We could pin to 0.58.1 or open a separate PR that fixes these warning but I don't want to mix it with this change.

iNecas:
The ideal would be: fix rubocop outside of this PR. After it's merged, rebase this one

orrabin:
@iNecas I opened: https://github.com/fog/fog-ovirt/pull/40

orrabin:
@iNecas I rebased this, tests are passing now

iNecas (approved):
Thanks @orrabin

PR #26, opened 2018-06-05
Gets correct cluster from datacenter

orrabin:
This fixes http://projects.theforeman.org/issues/23815

### DIFF
--- a/lib/fog/ovirt/requests/compute/v4/create_vm.rb
+++ b/lib/fog/ovirt/requests/compute/v4/create_vm.rb
@@ -27,9 +27,21 @@ module Fog
           # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
           def create_vm(attrs)
             attrs = attrs.dup
-            attrs[:cluster_name] ||= datacenter.clusters.first.name unless attrs[:cluster]
+
+            if attrs[:cluster].present?
+              attrs[:cluster] = client.system_service.clusters_service.cluster_service(attrs[:cluster]).get
+            else
+              if attrs[:cluster_name].present?
+                cluster = client.system_service.clusters_service.list(:search => "name=#{attrs[:cluster_name]}").first
+              else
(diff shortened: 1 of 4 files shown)


lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -27,9 +27,15 @@ def process_vm_opts(opts)
           # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
           def create_vm(attrs)
             attrs = attrs.dup
-            attrs[:cluster_name] ||= datacenter.clusters.first.name unless attrs[:cluster]
+
+            unless attrs[:cluster]
+              cluster = client.system_service.clusters_service.list(:search => "datacenter=#{datacenter_hash[:name]}").first
+              attrs[:cluster] = cluster
+              attrs[:cluster_name] ||= cluster.name
```
adamruzicka: Can there be a situation where `attrs[:cluster]` is unset, but `attrs[:cluster_name]` is set?

orrabin: Not that I've seen so far but it's possible if someone calls `create_vm` with the cluster name.
However this should probably change - if there is a name we should search by name and not datacenter and it we searched by datacenter the name should follow.

orrabin:
@adamruzicka thanks, I changed this to include an option for a provided `cluster_name` with no id.

adamruzicka (changes requested):
Looks good to me but :policeman: is not happy

orrabin:
@adamruzicka there were too many lines, I noticed we had an `if` with only 1 statement in it so it was a good opportunity to refactor that as well.

orrabin:
@adamruzicka rubocop was failing on an unrelated file this time, I'm fixing here but this will causes failures in other PRs as well.

adamruzicka (approved):

adamruzicka:
Thanks @orrabin !

PR #27, opened 2018-06-06
Fixing rubocop 0.57 errors

orrabin:
Looks like Style/MethodMissing was split into 2 cops:  Style/MissingRespondToMissing and Style/MethodMissingSuper

### DIFF
--- a/lib/fog/ovirt/compute.rb
+++ b/lib/fog/ovirt/compute.rb
@@ -33,7 +33,7 @@ module Fog
           @client = client
         end
 
-        # rubocop:disable Style/MethodMissing
+        # rubocop:disable Style/MethodMissingSuper, Style/MissingRespondToMissing
         def method_missing(symbol, *args)
           if block_given?
             @client.__send__(symbol, *args) do |*block_args|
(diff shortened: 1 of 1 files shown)


lib/fog/ovirt/compute.rb
```diff
@@ -33,7 +33,7 @@ def initialize(client)
           @client = client
         end
 
-        # rubocop:disable Style/MethodMissing
+        # rubocop:disable Style/MethodMissingSuper, Style/MissingRespondToMissing
```
ShimShtein, line 36: We can rename `#respond_to?` into `#respond_to_missing?`. This will actually fix the cop, and make our code better.
See https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding

ShimShtein (changes requested):
I prefer fixing `Style/MissingRespondToMissing`. `MethodMissingSuper` is OK though.

lib/fog/ovirt/compute.rb
orrabin: That was already a rule of `Style/MethodMissing` which we decided not to fix originally so I would suggest fixing rubocop errors now to unblock other PRs and any refactoring to be done later in another PR as that would also require more testing.

ShimShtein:
Merged. Thanks @orrabin.

I still think we need to fix this and not create technical debt for nothing, but I'll do it in a separate PR.

PR #29, opened 2018-06-22
Remove dependencies version constraint

### LINKED ISSUE #4000: Providers with pessimistic fog-core dependencies prevent bundling on edge
Still getting failures on Travis with `gemfiles/Gemfile-edge` after fixing `fog-json` in #3999

```
Resolving dependencies...
Bundler could not find compatible versions for gem "fog-core":
  In Gemfile-edge:
    fog was resolved to 2.0.0, which depends on
      fog-core (~> 1.45)

    fog was resolved to 2.0.0, which depends on
      fog-json was resolved to 1.2.0, which depends on
        fog-core
```

So master fog/fog@610c179 currently has `fog-core ~> 1.45` and fog-core/master fog/fog-core@7865ef7 declares itself as v2.1 (which bundler does not show and would have made this obvious).

So `fog` master is not currently compatible with `fog-core` master.

The fix is complicated as a number of providers also depend on `fog-core ~> 1.45` and are included by `fog`

I just released `fog-brightbox v0.15.1` which releases the version dependencies there so we are no longer one of the offending providers.

The providers that are preventing this from resolving are thus:

- [x] fog-aliyun
- [x] fog-cloudatcost
- [x] fog-digitalocean
- [ ] fog-joyent
- [x] fog-ovirt
- [x] fog-powerdns

plribeiro3000:
This is needed to help fix fog/fog#4000

### DIFF
--- a/fog-ovirt.gemspec
+++ b/fog-ovirt.gemspec
@@ -20,15 +20,15 @@ Gem::Specification.new do |spec|
 
   spec.require_paths = ["lib"]
 
-  spec.add_dependency("fog-core", "~> 1.45")
+  spec.add_dependency("fog-core")
   spec.add_dependency("fog-json")
-  spec.add_dependency("fog-xml", "~> 0.1.1")
+  spec.add_dependency("fog-xml")
   spec.add_dependency("ovirt-engine-sdk", ">= 4.1.3")
   spec.add_dependency("rbovirt", "~> 0.1.5")
 
-  spec.add_development_dependency "bundler", "~> 1.10"
-  spec.add_development_dependency "pry", "~> 0.10"
-  spec.add_development_dependency "rake", "~> 10.0"
+  spec.add_development_dependency "bundler"
+  spec.add_development_dependency "pry"
+  spec.add_development_dependency "rake"
   spec.add_development_dependency "rubocop", "~> 0.52"
-  spec.add_development_dependency "shindo", "~> 0.3"
+  spec.add_development_dependency "shindo"
 end


plribeiro3000:
@orrabin Due the urgency on this matter, i'm gonna merge it in and release. But feel free to comment your thoughts here if you want.

Thanks

orrabin:
@plribeiro3000 Foreman still uses those versions (https://github.com/theforeman/foreman/blob/develop/Gemfile#L34) and this can cause conflicts.
We are in the middle of fixes on the integration with api v4 that we need to get into a stable fog-ovirt version and can't bump the versions in Foreman at the moment.

I am fine with this change after we get out the version with the v4 fixes but unfortunately I will likely have to revert this to get that version ready and add again after it's released.

orrabin:
@plribeiro3000 I can avoid any other versions before we have all the critical fixes in and release 2 versions one after the other - fixes + dependency, then reapply your patch and release again.

plribeiro3000:
Hmmmm. Not sure i understood @orrabin 

The scenario we are trying to fix on `fog` is because we can't have the providers specifying its own `fog-json`, `fog-xml` and `fog-core` versions or `fog` wont install gracefully at some point.

Now, for foreman, the problem is that you need to relay on a stable version of `fog-ovirt`. Can't you just pinpoint `fog-ovirt` version to the one before the latest?

orrabin:
@plribeiro3000 ignore my previous comment, I thought you were going to force more advanced versions of `fog-core`. We need to release another stable version of `fog-ovirt` soon after a few PRs are merged that are urgent to foreman but if no version is required this should still be fine with foreman pinning `fog-ovirt`, `fog` and `fog-core`.

orrabin:
@plribeiro3000 thank you for fixing and releasing a version so fast!

orrabin:
@plribeiro3000 why was this merged without a PR: https://github.com/fog/fog-ovirt/commit/3a023e28e477c1c80f03f8387ee86a1397b49eac ? This does not look related to the issue above and `fog-ovirt` is currently not loading with this commit.

plribeiro3000:
@orrabin Ouch. sorry about not opening a PR.

It is not related to this issue but is related to another one fog/fog#3430

The `core` file was something from `fog` before the extraction. the related code should be on the main file and use autoload to avoid loading all files at boot time.

Why is is not loading anymore? Do you get an error?

plribeiro3000:
I just checked and i can load `bundle console` locally.

Maybe i'm doing something wrong?

orrabin:
@plribeiro3000 I haven't found the error yet but when I try to load providers in Foreman from: https://github.com/fog/fog-core/blob/v1.45.0/lib/fog/core/services_mixin.rb#L22 ` Fog.services` does not include ovirt.
Please notice I need to get it working with `fog-core` 1.45 and `fog` 1.42.

plribeiro3000:
Hmmmm. I tried and got this

```ruby
2.4.1 :007 > Fog.services
 => {:compute=>[:ovirt]}
```

Maybe foreman is requiring that `core` file manually?

Because the only real difference is that the code is now defined in another file, which is gonna be loaded first.

orrabin:
@plribeiro3000 when loading from Foreman it doesn't reach `lib/fog/ovirt.rb`, the first thing to load is `lib/fog/ovirt/compute.rb`.

This looks like it's the result to this in fog: https://github.com/fog/fog/blob/master/lib/fog/ovirt.rb ? This is probably a leftover that should have been deleted when creating `fog-ovirt`.

@plribeiro3000 even after fixing this in fog since we still need version 1.42 - is this something that can be backported?
If not, I need to revert your change. Please open a PR in that case which I'll add after there is a version of fog with the fix.

plribeiro3000:
Yes, this was something to be fixed here.

The main file on the repo `lib/fog/ovirt` is gonna be loaded automatically by bundler.
This means that at bundler load it will define the main namespace `Fog::Ovirt` and others like `Fog::Compute::Ovirt` without loading the entire stack.

This is needed to fix the loading time issue fog/fog#3430. This is something i fixed some time ago but `fog-ovirt` slipped from my radar somehow.

What i'm saying is: This is something to be fixed here rather somewhere else. If Foreman is loading the file `lib/fog/ovirt/compute` directly it need to change to load the main file instead: `lib/fog/ovirt`

I think i did not get your point on backporting this or even undo. Whats the problem with Foreman?

orrabin:
@plribeiro3000 fog/fog is doing this, not foreman: https://github.com/fog/fog/blob/master/lib/fog/ovirt.rb.
This is a leftover from the plugin extraction that needs to be fixed there, otherwise if fog/fog is loaded before fog/ovirt it will load in the wrong order.

I changed locally in fog/fog and ran from local code instead of from gem and it loaded in the correct order.

iNecas:
@plribeiro3000 the problem is Foreman's current stable depends on fog 1.42 and we would like to keep the fog-ovirt compatible with this version for some time (as much as we like keeping pace with upstream, given we rely on multiple providers, it's not feasible to move to new version every day, especially when in preparation for new upstream release - we have 1.18 RC1 at the moment).

Therefore, to make the maintenance easier and the work more effective (which I believe both Fog and Foreman community benefits from), we should try to find solution that doesn't require the master branch of fog-ovirt to require on fog > 1.42. Otherwise, the time that we could spend on making fog-ovirt better, we will will spend on cherry-picking every single bugfix to APIv4 intergration to older branch.

plribeiro3000:
@orrabin Ouch. We need to remove it then. Could you push a PR?

@iNecas Got. The thing is that nothing needs to change here then. Just pinpoint the version of `fog-core` and `fog-ovirt` in Foreman.

The point you mentioned is the exact reason for those changes. The dependency tree go up to 5 levels on `fog`. Since we have several providers, if each require a specific version we will always have conflicts so we need to push for a change.

This means, the version to be used need to be pinpointed by the final user or in this case, `foreman` itself.

orrabin:
@plribeiro3000 :+1:

iNecas:
Our interest is to pin `fog-core` but not pinning `fog-ovirt` right now, and I understand your issues as well. Hopefully, we will find a way how to make everyone happy here.

Since it seems the main issue is the `require 'fog/ovirt/compute'` in fog, I wonder if keeping this one file in fog-ovirt for compatibility reasons couldn't give us the benefits of working both with latest-greatest fog, as well as with 1.42?

plribeiro3000:
@iNecas I'm not sure i'm following you.

Which file are you talking about?

iNecas:
Sry, scratch my last comment about `'fog/ovirt/compute'` I read the comments wrong, but @orrabin mentioned she has some ideas to move this forward, so leaving this up to her, as she's obviously more knowledgeable about it.

plribeiro3000:
@iNecas Coolio

@orrabin If we do remove that file from `fog` and do a new release does it fix the problem for you?

Do we need to release a bugfix version for the `1.x` series or can we move on with `2.x`?

orrabin:
@plribeiro3000 for 1.x: https://github.com/fog/fog/pull/4002
I'll create another one for master to remove 2 more leftover files.
Edit: and yes this fixes the issue, I tested locally with my PR.
Thanks for creating the branch for me!

orrabin:
for `2.x`: https://github.com/fog/fog/pull/4003

ekohl:
I do wonder why the constraint was removed completely. Why not change `~>` to `>=` so you at least maintain the lower bound?

plribeiro3000:
@ekohl Because even the `~>` can have clash changes if one provider depends on `~> 1.1`and another on `~> 1.2`.

Do you have any problems with this? 

Maybe a collateral damage?

ekohl:
@plribeiro3000 that's not what I said. I said that we should probably change to `>= 1.45.0` because I doubt this gem suddenly works with `1.44.0`. It will still allow `2.0` like this change. While unlikely, it's good to be explicit about lower bounds.

plribeiro3000:
Ouch. Did not read it correctly.

Sorry for the confusion.

The main point when we started pushing this to all providers is that the final user can define the version or leave it open and use the latest.

I do agree with you on the lower bound but we might have another provider that does depend on a lower bound and then again we do have the same problem.

Maybe we can push a DSL on `fog-core` where each provider can define its version dependency?

ekohl:
A DSL sounds wrong. Ruby gems can indicate dependencies and a bundle install should never allow a situation where it's broken out of the box. I appreciate this is hard, but IMHO if we know something is required because of incompatible APIs we should declare this. Upper bounds are always a bit tricky because you don't know. `~>` is defensive and makes major version bumps painful. A lot of Linux packaging has been using `>=` which is open which can make a lot of sense because you want to allow dependencies to be as wide as possible.

tl;dr: I still think we should specify the lower bounds, I'm neutral about upper bounds.

plribeiro3000:
@ekohl You are right. There is no sense in doing what i suggested. 😅

I believe we can do a lower bound here if it is not a newer version.

Which version do you think that make sense here?

ekohl:
I'm no expert on the code so I'd suggest restoring the previous lower bound.

plribeiro3000:
Which would be `1.45` for `fog-core`

I think we should be fine with it. It is almost 1 year old now.

PR #28, opened 2018-06-14
Fixed method_missing cops pattern

ShimShtein:
Changed `#respond_to?` to `#respond_to_missing?`
Changed both `method_missing` and `respond_to_missing` to call to `super`, if nothing is found on the `@client`

PR #30, opened 2018-06-27
Fix rubocop errors

shiramax:


shiramax:
@adamruzicka can you please review?

adamruzicka:
Rubocop is still failing here, maybe something like https://github.com/fog/fog-ovirt/pull/28 would be the way to go?

PR #34, opened 2018-07-22
Add OS boot order when template is empty

orrabin:


lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -65,6 +66,21 @@ def create_vm(attrs)
             vms_service.add(new_vm)
           end
           # rubocop:enable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+
+          # rubocop:disable Metrics/AbcSize
+          def update_os_attrs(attrs)
+            attrs[:os] ||= {}
+            attrs[:os][:type] ||= attrs[:os_type] || "unassigned"
```
ShimShtein: nitpick: you can use `#fetch` here too.

ShimShtein: Also: shouldn't we `delete` attributes that are not part of SDK's schema?

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -65,6 +66,21 @@ def create_vm(attrs)
...
+            attrs[:os][:boot] ||= [attrs.fetch(:boot_dev1, OvirtSDK4::BootDevice::NETWORK), attrs.fetch(:boot_dev2, OvirtSDK4::BootDevice::HD)]
+            if attrs[:first_boot_dev]
+              attrs[:os][:boot] = attrs[:os][:boot].sort_by.with_index do |device, index|
```
ShimShtein: Two things:
1. We can use [`#find_index`](https://ruby-doc.org/core-2.2.0/Array.html#method-i-find_index) instead of `sort_by.with_index`
2. Do we expect the user to pass `attrs[:os][:boot]` list? I think it would be empty otherwise.

ShimShtein (changes requested):
A couple of small changes please.

lib/fog/ovirt/requests/compute/v4/create_vm.rb
orrabin: 1. I'll change it
2. we fill it on liמe 74

lib/fog/ovirt/requests/compute/v4/create_vm.rb
orrabin: we don't have to delete them but if we do, maybe instead of using fetch it should be:
` attrs.delete(:os_type) || "unassigned"` ?

lib/fog/ovirt/requests/compute/v4/create_vm.rb
orrabin: find_index does not do what I want, this reorders the list so `first_boot_dev` is first

orrabin:
@ShimShtein I updated the PR, thanks for the comments.

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -65,6 +66,21 @@ def create_vm(attrs)
...
           end
           # rubocop:enable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+
+          # rubocop:disable Metrics/AbcSize
+          def update_os_attrs(attrs)
+            attrs[:os] ||= {}
+            attrs[:os][:type] ||= "Other OS"
+            attrs[:os][:boot] ||= [attrs.fetch(:boot_dev1, OvirtSDK4::BootDevice::NETWORK), attrs.fetch(:boot_dev2, OvirtSDK4::BootDevice::HD)]
+            if attrs["first_boot_dev"]
+              attrs[:os][:boot] = attrs[:os][:boot].sort_by.with_index do |device, index|
```
ShimShtein: I think this looks a bit better:
``` ruby
# put first_boot_dev first
attrs[:os][:boot] = [attrs["first_boot_dev"]] + (attrs[:os][:boot] - [attrs["first_boot_dev"]])
```

or even:

``` ruby
# put first_boot_dev first
attrs[:os][:boot] -= [attrs["first_boot_dev"]]
attrs[:os][:boot] = [attrs["first_boot_dev"]] + attrs[:os][:boot]
```

ShimShtein: Found an option with less brackets:

``` ruby
attrs[:os][:boot] = attrs[:os][:boot].without(attrs[:first_boot_dev]).prepend(attrs[:first_boot_dev])
```

orrabin:
@ShimShtein Thanks for the comment, I changed the boot device array.

ShimShtein:
@orrabin, do we need to handle also `#update_vm`?

orrabin:
@ShimShtein I can add it to `update_vm` but right now from foreman we don't change the boot order on update.

ShimShtein (approved):

ShimShtein:
Merged. Thanks @orrabin!

PR #35, opened 2018-08-13
Fixes #24564 - Clusters and Vms are shown by selected DC

shiramax:


lib/fog/ovirt/requests/compute/v4/list_clusters.rb
```diff
@@ -3,8 +3,9 @@ module Compute
     class Ovirt
       class V4
         class Real
-          def list_clusters(filters = {})
-            client.system_service.clusters_service.list(filters).map { |ovirt_obj| ovirt_attrs ovirt_obj }
+          def list_clusters(opts = {})
+            opts[:search] = "datacenter=#{datacenter_hash[:name]}"
```
ShimShtein: If `opts[:search]` already contained some filters, they would be overridden. Is it intentional?

Example:

``` ruby
opts = { search: 'tag = something' }
client.list_clusters(opts)
# => will return all clusters, event without the tag
```

lib/fog/ovirt/requests/compute/v4/list_virtual_machines.rb
```diff
@@ -7,9 +7,9 @@ def list_virtual_machines(filters = {})
             filters = filters.dup
             page = filters.delete(:page)
             without_details = filters.delete(:without_details)
-
             filters[:all_content] = true unless without_details
             filters[:search] ||= ""
+            filters[:search] += "datacenter=#{datacenter_hash[:name]}"
```
ShimShtein: if I already have `foo=bar` in search field, it will end up with `foo=bardatacenter=bleh`, which is not correct.

ShimShtein (changes requested):

shiramax:
@ShimShtein could you please review again :) ?

lib/fog/ovirt/compute/v4.rb
```diff
@@ -167,6 +166,7 @@ def datacenter
           end
 
           def datacenter_hash
+            @datacenter_hash ||= (datacenters.select { |x| x[:id] == datacenter })[0] if datacenter
```
ShimShtein: You can use `#find` instead of `select()[0]`.

lib/fog/ovirt/requests/compute/v4/list_clusters.rb
```diff
@@ -3,8 +3,9 @@ module Compute
     class Ovirt
       class V4
         class Real
-          def list_clusters(filters = {})
-            client.system_service.clusters_service.list(filters).map { |ovirt_obj| ovirt_attrs ovirt_obj }
+          def list_clusters(opts = {})
+            opts[:search] += " datacenter=#{datacenter_hash[:name]}"
+            client.system_service.clusters_service.list(opts).map { |ovirt_obj| ovirt_attrs ovirt_obj }
```
ShimShtein: Nitpick: can we put `()` around `ovirt_obj`? `ovirt_attrs(ovirt_obj)` is a bit easier to read.

lib/fog/ovirt/requests/compute/v4/list_clusters.rb
```diff
@@ -3,8 +3,9 @@ module Compute
     class Ovirt
       class V4
         class Real
-          def list_clusters(filters = {})
-            client.system_service.clusters_service.list(filters).map { |ovirt_obj| ovirt_attrs ovirt_obj }
+          def list_clusters(opts = {})
+            opts[:search] += " datacenter=#{datacenter_hash[:name]}"
```
ShimShtein: This will error if `opts[:search]` is not defined.

lib/fog/ovirt/requests/compute/v4/list_virtual_machines.rb
```diff
@@ -3,16 +3,18 @@ module Compute
     class Ovirt
       class V4
         class Real
+          # rubocop:disable Metrics/AbcSize
```
ShimShtein: Instead of disabling this metric, could we just refactor all `filters` assignments into a private method?

ShimShtein (changes requested):

shiramax:
@ShimShtein let me know if its ok now :)

lib/fog/ovirt/compute/v4.rb
```diff
@@ -167,13 +166,21 @@ def datacenter
...
             @datacenter_hash ||= datacenters.first
           end
 
           def blank_template
             @blank_template ||= client.system_service.get.special_objects.blank_template
           end
 
+          def create_search_by_datacenter(search, page)
+            search ||= ""
+            search += " datacenter=#{datacenter_hash[:name]}"
```
ShimShtein: Nitpick: why not passing datacenter name as a parameter? This way you will not have dependencies on state, only on input parameters.

lib/fog/ovirt/requests/compute/v4/list_clusters.rb
```diff
@@ -3,8 +3,10 @@ module Compute
     class Ovirt
       class V4
         class Real
-          def list_clusters(filters = {})
-            client.system_service.clusters_service.list(filters).map { |ovirt_obj| ovirt_attrs ovirt_obj }
+          def list_clusters(opts = {})
+            opts[:search] ||= ""
```
ShimShtein: You don't need this anymore.

orrabin: This is already done in `create_search_by_datacenter`

lib/fog/ovirt/compute/v4.rb
```diff
@@ -167,13 +166,21 @@ def datacenter
...
           def datacenter_hash
+            @datacenter_hash ||= datacenters.find { |x| x[:id] == datacenter } if datacenter
             @datacenter_hash ||= datacenters.first
           end
 
           def blank_template
             @blank_template ||= client.system_service.get.special_objects.blank_template
           end
 
+          def create_search_by_datacenter(search, page)
```
ShimShtein: If you don't always have values for page or datacenter name, you can make them [named parameters](https://robots.thoughtbot.com/ruby-2-keyword-arguments), or at least put defaults, so you will not have to pass `nil` if you don't have something to put there.

ShimShtein (changes requested):
Almost there, a couple of small changes.

shiramax:
@ShimShtein thanks! :) should be ready now :)

ShimShtein (approved):

ShimShtein:
Merged, thanks @shiramax !

PR #40, opened 2018-10-23
Fix rubocop 0.59 errors

orrabin:


iNecas:
Well, I feel quite strong against applying this rule. I think it actually hurts the readability. I understand there were good intentions when someone introduced this rule, but sometimes, the good intentions are just not enough :)

orrabin:
I actually liked that one but it's a matter of personal preference and I agree it should stay that way so I'm disabling the cop.

orrabin:
@iNecas could you take another look please?

iNecas:
That looks better to me :) Thanks @orrabin

iNecas (approved):

PR #43, opened 2018-11-07
Fix Style/IfUnlessModifier and Layout/AlignHash

adamruzicka:


lib/fog/ovirt/compute/v4.rb
```diff
@@ -63,9 +63,7 @@ def ovirt_attrs(obj)
               # ignore nil values
               next if value.nil?
 
-              if key != :data_center && value.respond_to?(:href) && value.href && value.respond_to?(:id)
-                opts[key] = value.id
-              end
+              opts[key] = value.id if key != :data_center && value.respond_to?(:href) && value.href && value.respond_to?(:id)
```
orrabin, line 66: The cop in general is correct but in this one case it makes it even harder to read with all the `and` conditions.
Commenting here as a reminder to refactor this in a separate PR.

orrabin (approved):

orrabin:
Thanks @adamruzicka!

PR #45, opened 2018-11-20
Refs #20394 - Send the volume parameters in template

shiramax:


lib/fog/ovirt/requests/compute/v4/list_template_volumes.rb
```diff
@@ -3,15 +3,22 @@ module Compute
...
-
             attachments.map do |attachment|
-              ovirt_attrs client.follow_link(attachment.disk)
+              attachment_disk = client.follow_link(attachment.disk)
+              attachment = client.follow_link(attachment)
+              bootable = attachment.bootable
+              interface = attachment.interface
+              attachment_disk.bootable = bootable if attachment_disk.bootable.nil?
+              attachment_disk.interface = interface if attachment_disk.interface.nil?
+              attachment_disk.storage_domain = attachment_disk.storage_domains[0].id
```
orrabin: This is probably not possible but I would suggest checking there is something in storage_domains before trying to take the first one.

orrabin (changes requested):
Works well, I have 1 nitpick, after that this can be merged.

shiramax:
Thanks @orrabin , please review again :)

orrabin (approved):
The UI now shows the storage domain of the disk + the first on the list.
This is consistent with v3 that has the same bug. It's fixed after save and it's an issue in foreman UI.

orrabin:
Thanks @shiramax !

PR #41, opened 2018-10-24
Fixes #24808 - Fix host boolean values in update

shiramax:
Updating a Host caused the parameters - Stateless, Start in Pause Mode, Delete Protection and more to be checked.

PR #46, opened 2018-11-21
Fixes #14833 - add template cloning for create_vm

orrabin:


lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -9,18 +9,48 @@ def check_for_option(opts, name)
...
-            # Create disks map
-            opts[:disks] = template_disks.collect { |d| { :id => d.id, :storagedomain => storagedomain_id } }
+
+            # Create disks map "clone" from disks, rather then template
+            if opts[:disks] && !opts[:disks].empty?
+              disk_attachments = []
+              opts[:disks].each do |disk|
+                active = disk["active"].present? ? disk["active"] == "true" : true
+                sparse = disk["sparse"].present? ? disk["sparse"] == "true" : true
+                interface = disk["interface"] || OvirtSDK4::DiskInterface::VIRTIO
```
orrabin: @iNecas we now know that the sdk can't change disk interface and provisioned size in the same api call as the cloning because the ovirt api does not expect this.
Do you think I should delete them for now or comment out with a comment?
There is no RFE to add this to the api for now but a comment with the RFE link if one is opened cold be helpful.

orrabin:
@ShimShtein I'm working on the PR on the Foreman side so this can be tested properly.

lib/fog/ovirt/requests/compute/v4/create_vm.rb
iNecas: +1 for creating and RFE and linking from here

orrabin:
@iNecas the interface and size are actually not part of create in ovirt, they are only allowed to be changed on update.
There should be another fix to update_vm after this is merged and no need for RFE for ovirt.

lib/fog/ovirt/compute/v4.rb
```diff
@@ -60,6 +60,11 @@ def ovirt_attrs(obj)
                 next
               end
 
+              if key == :storage_domain && obj.is_a?(OvirtSDK4::Disk) && obj.storage_domains.size == 1
```
orrabin: This is because the storage domain did not show up in foreman so cloning would send the first in the list of possible storage domains.

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -9,18 +9,40 @@ def check_for_option(opts, name)
...
             opts[:clone] = true unless opts[:clone] == true || template_disks.empty? || template_disks.all? { |d| d.storage_domain == storagedomain_id }
 
-            # Create disks map
-            opts[:disks] = template_disks.collect { |d| { :id => d.id, :storagedomain => storagedomain_id } }
+
+            # Create disks map "clone" from disks, rather then template
+            if opts[:disks] && !opts[:disks].empty?
+              disk_attachments = []
+              opts[:disks].each do |disk|
+                sparse = disk["sparse"].present? ? disk["sparse"] == "true" : true
```
shiramax: Since the PR https://github.com/fog/fog-ovirt/pull/41 was merged, the Boolean values should stay Boolean and not string.

shiramax (changes requested):

orrabin:
This can be tested with: https://github.com/theforeman/foreman/pull/6295

This will need rebase after: https://github.com/fog/fog-ovirt/pull/45 is merged.

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -3,24 +3,25 @@ module Compute
...
         class Real
-          def check_for_option(opts, name)
-            opts[name.to_sym] || opts[(name + "_name").to_sym]
-          end
 
           # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
           def process_vm_opts(opts)
-            return unless check_for_option(opts, "template") && check_for_option(opts, "storagedomain")
+            disk_attachments = []
+            opts[:disks].each do |disk_from_attachment|
```
ShimShtein: You can use `opts[:disks].map` to create disks list.

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -3,24 +3,25 @@ module Compute
...
+              disk = {
+                  :id => disk_from_attachment["id"],
+                  :format => disk_from_attachment["format"]|| OvirtSDK4::DiskFormat::COW,
+                  :sparse => sparse,
+                  :storage_domains => [client.system_service.storage_domains_service.storage_domain_service(disk_from_attachment["storage_domain"]).get]
+              }
 
-            # Make sure the 'clone' option is set if any of the disks defined by
-            # the template is stored on a different storage domain than requested
-            opts[:clone] = true unless opts[:clone] == true || template_disks.empty? || template_disks.all? { |d| d.storage_domain == storagedomain_id }
```
ShimShtein, line 20: We are not setting `clone` attribute when disks are in different storage domains anymore. Is it intentional?

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -59,11 +60,13 @@ def create_vm(attrs)
             end
             attrs[:memory_policy] = OvirtSDK4::MemoryPolicy.new(:guaranteed => attrs[:memory]) if attrs[:memory].to_i < Fog::Compute::Ovirt::DISK_SIZE_TO_GB
             attrs[:high_availability] = OvirtSDK4::HighAvailability.new(:enabled => attrs[:ha] == "1") if attrs[:ha].present?
+            
+            if attrs[:clone] == true && attrs[:disks].present?
```
ShimShtein: Is there an option where `:clone` is set to `false`?

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -3,24 +3,25 @@ module Compute
...
-            # Make sure the 'clone' option is set if any of the disks defined by
-            # the template is stored on a different storage domain than requested
-            opts[:clone] = true unless opts[:clone] == true || template_disks.empty? || template_disks.all? { |d| d.storage_domain == storagedomain_id }
+                disk_attachments << OvirtSDK4::DiskAttachment.new(:disk => disk)
+              end
+            opts[:disk_attachments] = disk_attachments
 
-            # Create disks map
-            opts[:disks] = template_disks.collect { |d| { :id => d.id, :storagedomain => storagedomain_id } }
+            opts.delete(:disks)
```
ShimShtein, line 21: The function will return the contents of `opts[:disks]`. I think it would be better to return the whole `opts` hash.

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -3,24 +3,25 @@ module Compute
     class Ovirt
       class V4
         class Real
-          def check_for_option(opts, name)
-            opts[name.to_sym] || opts[(name + "_name").to_sym]
-          end
 
           # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
```
ShimShtein: With all those changes, the method looks quite slim. Can we readd those rubocop rules back?

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -59,11 +60,13 @@ def create_vm(attrs)
...
+            
+            if attrs[:clone] == true && attrs[:disks].present?
+              process_vm_opts(attrs)
+            end
 
-            # TODO: handle cloning from template
-            process_vm_opts(attrs)
             new_vm = OvirtSDK4::Vm.new(attrs)
-            vms_service.add(new_vm)
+            vms_service.add(new_vm, :clone => attrs[:clone])
```
ShimShtein, line 64: I see that `:clone` is passed as part of attrs. It feels redundant to send it both in `attrs` and as a parameter here.

ShimShtein (changes requested):

lib/fog/ovirt/requests/compute/v4/create_vm.rb
orrabin: Yes, a template can have disks from different storage domains, we do not have 1 storage domain per template then we compare to.
We clone disks that have changed from the original storage domain they had in the template and that is checked before we send them to fog-ovirt.

lib/fog/ovirt/requests/compute/v4/create_vm.rb
orrabin: It can be nil, I haven't sent `false` there through foreman but it can theoretically be false.

lib/fog/ovirt/requests/compute/v4/create_vm.rb
orrabin: We have to send it here, that is how to clone according to the documentation.
It's not needed in attrs and disregarded by the sdk but that is where we receive it.
I could change this to `:clone => attrs.delete(:clone)` but there are other fields the sdk doesn't care about and we don't remove all of them.

lib/fog/ovirt/requests/compute/v4/create_vm.rb
orrabin: No need to return anything here, we don't use the return value, just change the opts which are the attrs from `create_vm`.

lib/fog/ovirt/requests/compute/v4/create_vm.rb
orrabin: Yes, at least the size, I need to check the others.

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -3,26 +3,23 @@ module Compute
...
         class Real
-          def check_for_option(opts, name)
-            opts[name.to_sym] || opts[(name + "_name").to_sym]
-          end
-
-          # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
-          def process_vm_opts(opts)
-            return unless check_for_option(opts, "template") && check_for_option(opts, "storagedomain")
+          def create_disk_attachment_from_disk(disk_to_attachment)
+            sparse = disk_to_attachment["sparse"].present? ? disk_to_attachment["sparse"] : true
```
ShimShtein: We can use `fetch` with default here.
```suggestion
            sparse = disk_to_attachment.fetch("sparse", true)
```

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -3,26 +3,23 @@ module Compute
...
 
-            template_id = opts[:template] || client.system_service.templates_service.search(:name => opts[:template_name]).first.id
-            template_disks = client.system_service.templates_service.template_service(template_id).get.disk_attachments
-            storagedomain_id = opts[:storagedomain] || storagedomains.select { |s| s.name == opts[:storagedomain_name] }.first.id
+            disk = {
+              :id => disk_to_attachment["id"],
+              :format => disk_to_attachment["format"] || OvirtSDK4::DiskFormat::COW,
```
ShimShtein: Ditto `fetch`

```suggestion
              :format => disk_to_attachment.fetch("format", OvirtSDK4::DiskFormat::COW),
```

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -3,26 +3,23 @@ module Compute
...
+              :sparse => sparse,
+              :storage_domains => [client.system_service.storage_domains_service.storage_domain_service(disk_to_attachment["storage_domain"]).get]
```
ShimShtein: Can you please separate this to two lines?

``` ruby
storage_domains = client.system_service.storage_domains_service.storage_domain_service(disk_to_attachment["storage_domain"]).get

disk = {
# ...
:storage_domains => storage_domains
}
```

lib/fog/ovirt/requests/compute/v4/create_vm.rb
```diff
@@ -3,26 +3,23 @@ module Compute
...
 
-            # Make sure the 'clone' option is set if any of the disks defined by
-            # the template is stored on a different storage domain than requested
-            opts[:clone] = true unless opts[:clone] == true || template_disks.empty? || template_disks.all? { |d| d.storage_domain == storagedomain_id }
+            OvirtSDK4::DiskAttachment.new(:disk => disk)
+          end
 
-            # Create disks map
-            opts[:disks] = template_disks.collect { |d| { :id => d.id, :storagedomain => storagedomain_id } }
+          def process_vm_opts(opts)
```
ShimShtein: We can probably rename this to `process_vm_disks` (a bit less generic)

ShimShtein (changes requested):

orrabin:
@ShimShtein I rebased and fixed the remaining comments, please review again.

ShimShtein (approved):

ShimShtein:
Merged, thanks @orrabin !

PR #22, opened 2018-04-30
Enable ha flag when creating ovirt VM

shiramax:

PR #47, opened 2018-11-29
Fixes #24532 - Add support for setting keyboard layout

UXabre:
This allows us to set the VNC keyboard layout from within Foreman host creation; which, in case of VNC + non-US keyboard is mandatory IMO).

### DIFF
--- a/lib/fog/ovirt/compute/v4.rb
+++ b/lib/fog/ovirt/compute/v4.rb
@@ -95,6 +95,7 @@ module Fog
                 :address => value.address,
                 :port => value.port,
                 :secure_port => value.secure_port,
+                :keyboard_layout => value.keyboard_layout,
                 :subject => subject,
                 :monitors => value.monitors
               }


UXabre:
I noticed that the CC is to high, so I'll have to refactor this one out. However, I don't know how to refactor the class length?

orrabin:
@UXabre the failing cops can be disabled for `update_vm` and `create_vm`, they need some refactoring already without this change.

orrabin:
@UXabre thanks for opening this.
The code looks good, I do have one question - why is the keyboard layout it's own attribute in instance type? in't it part of the display?

I'll wait for @shiramax to also take a look as she has been working on improvements of our console display.

UXabre:
Hi @orrabin,
You're correct that keyboard_layout is part of the display object, but, although I was able to read-out the keyboard layout from within Foreman (but for this I also had to make a change to this plugin); I wasn't able to change the keyboard setting while creating a new host. 
The reason is that, as far as I can tell, I'm not able to link an input field in Foreman to a nested variable, in this case display[:keyboard_layout] (full disclosure, I did manage to read-out the variable but never to change it). That's why I started looking at the addition of the HA setting from within foreman, as it seems to be fairly similar. 

fog patch for HA:
https://github.com/fog/fog-ovirt/pull/22/files
Foreman patch for HA:
https://github.com/theforeman/foreman/pull/5513

shiramax:
@UXabre thank you for you contribution! 

I'm not sure the right approach here is to set the type of the console in the compute resource, because we need to mimic oVirt behavior as much possible, and in ovirt it possible to change console type in each VM and not per DC. 
On the other hand we don't want  to put the entire options of ovirt in foreman, 
So my original plan for the issue: [24532](https://projects.theforeman.org/issues/24532)
was to set by default vnc+spice console and , then each user will be able to choose in ovirt the console he want to use (please see the BZ as well - [1468742](https://bugzilla.redhat.com/show_bug.cgi?id=1468742) )

Do you think the solution of SPICE+VNC will be enough for you ?

UXabre:
Hi @shiramax, I've used the same approach here as was done for the libVirt plugin (just to have same experience there), however, I agree that this is actually very related to a specific VM. Originally my plan was to allow this choice to act as a default setting (to be in line with libVirt) but, in a later stage allow both libVirt as oVirt to have the choice at VM creation time instead; defaulting to whatever has been configured in the compute resource.

VNC+Spice works only well from within oVirt, because in oVirt you can select the console type at runtime. This is (currently) not supported in foreman and it seems to default to Spice instead. However, the html5 spice implementation seems a bit rough on the edges from time to time. Hence the choice for going with VNC **or** Spice to help clarify this to Foreman. **Perhaps, this could be optimized in the future to support this at runtime as well**, would be a nice addition IMO

As for the keyboard, this is, in either case, mandatory for VNC to work correctly. In Belgium, for instance, we have this French-alike keyboard layout, called fr-be in oVirt. If this defaults to en-us, then typing becomes a living hell, let alone, some keys don't even work :-( As this is machine specific, I want to allow, at VM creation time, to set this keyboard layout. 

But don't get me wrong, I also don't believe each and every oVirt option should be exposed to foreman. But in this case I believe it makes sense for streamlining the UX: User creates Console (via any which way) in foreman -> user connects to Console via either noVNC or some client -> user can type as expected. If we don't do this, an extra step of logging into oVirt would be required with an additional reboot before the settings would be applied.

I hope this clarifies my need for the keyboard setting a bit more :-)

_Just to be complete: in my company we want to take over a VM via the web console. This to streamline our support flow. Here, web Spice doesn't fulfill its promise yet so we have to change the console type everytime. However, most are not allowed to fiddle with oVirt settings, usually I would be contacted to change this; I don't want to be my company's bottleneck :p _

shiramax:
@UXabre thanks for the response, it clarifies your need for the keyboard. 
What do you think about creating a template in oVirt with the console configured as you wish, and then creating a host from foreman with that template?  Does this solution can meet your needs?

UXabre:
Hey @shiramax , thank you for bearing with me on this one and providing me this work-around :-)

AFAIK this looks like I'd have to create a few templates for each permutation of keyboard combinations with my set of existing templates? (we're an international company, so would be nice if I could keep one template for a certain type of application, withouth having to make for instance a NetworkController-en_us, NetworkControl-fr_be, NetworkController-jp, ...)

Anyway, perhaps this use-case that'd need more discussion in the form of a RFC on the Foreman Community site? I can imagine I'm not the only one having such a need but perhaps for various other reasons then I could possibly fathom. But that could just be an assumption of mine as well. Either way I think it will be educational for me to better understand the pros and the cons of having this feature in :-)

shiramax:
@UXabre  
It's a good idea to open RFC and see if other users are dealing with the same issue :)

UXabre:
Hi everyone, long time no see but I've managed to greatly minimize the impact on this fog plugin.
Only one variable is added in order to be able to read out the currently set keyboard!

shiramax (changes requested):
@UXabre thanks!
We need to connect between your PR and a red mine issue, by creating the title as following: 
"Fixes #xxxx - Add support for setting keyboard layout"
If you don't have a redmine issue please open one.

UXabre:
Hey @shiramax , can I link it to this one? https://projects.theforeman.org/issues/24532 or do I need to create a new one for fog-ovirt specifically?

Kind regards,
Arend

shiramax:
@UXabre , 
This issue fine, no need to open a new one :)  
But, What about adding spice console/vnc console?
here an example of how to add that: 
https://github.com/oVirt/ovirt-engine-sdk-ruby/blob/master/sdk/examples/add_vnc_console.rb

UXabre:
Hey @shiramax 

> But, What about adding spice console/vnc console?

I'm not sure I understand the question? This can already be set using the display.type field; it's what I use in the following Foreman PR: https://github.com/theforeman/foreman/pull/6555

Does that answer your question? 🙂

shiramax (approved):

PR #49, opened 2018-12-07
Fixes #25644 - Ticket retrieval returned ticket object instead of ticket value

UXabre:
As the object is returned instead of string, the console never really gets authenticated and thus this results in a "permission denied".


shiramax:
thanks, @UXabre, 
I've tested it and it's ready to merge @orrabin

orrabin (approved):

orrabin:
Thanks @UXabre !

PR #52, opened 2018-12-09
Ref #20394 - Fix sparse value

shiramax:


lib/fog/ovirt/requests/compute/v4/add_volume.rb
```diff
@@ -27,13 +27,14 @@ def add_options_defaults(options)
             options[:type] ||= OvirtSDK4::DiskType::DATA
             options[:format] ||= OvirtSDK4::DiskFormat::COW
 
-            options[:sparse] = true unless options[:sparse].present?
+            # options[:sparse] = true unless options[:sparse].present?
```
ShimShtein, line 30: Please don't leave commented out code. If it's not needed anymore, you can delete it.

lib/fog/ovirt/requests/compute/v4/add_volume.rb
```diff
@@ -27,13 +27,14 @@ def add_options_defaults(options)
...
             options[:quota] = options[:quota].present? ? client.system_service.data_centers_service.data_center_service(datacenter).quotas_service.quota_service(options[:quota]).get : nil
 
             options[:disk] ||= {}
             options[:disk][:storage_domains] ||= [client.system_service.storage_domains_service.storage_domain_service(options[:storage_domain_id]).get]
             options[:disk][:provisioned_size] ||= options.delete(:provisioned_size)
             options[:disk][:format] ||= options.delete(:format)
+
```
ShimShtein: Why the empty line?

ShimShtein (changes requested):
The commit seems OK, but I don't see its relevance to the main issue.
As far as I understand, it was an unused option that was set on vm level. It shouldn't affect the resulting VM at all.

shiramax:
thanks @ShimShtein , 
it's not an unused option, but its name can be misleading in foreman it's called preallocated.

lib/fog/ovirt/compute/v4.rb
```diff
@@ -51,7 +51,7 @@ def ovirt_attrs(obj)
               value = obj.instance_variable_get(v)
 
               if key == :network
-                opts[key] = client.follow_link(obj.vnic_profile).network.id
+                opts[key] = obj.vnic_profile.present? ? client.follow_link(obj.vnic_profile).network.id : value
```
adamruzicka, line 54: This is identical to https://github.com/fog/fog-ovirt/pull/39

lib/fog/ovirt/requests/compute/v4/add_volume.rb
ShimShtein: This changes the behaviour of line 37:
```ruby
 options[:disk][:sparse] ||= options.delete(:sparse)
```

Now if the option is not set, it will pass `nil` to ovirt, as opposed to `true` that was set by the commented out line.

ShimShtein (changes requested):

ShimShtein:
I think all the defaults should move to the corresponding `.delete` lines:
```ruby
            options[:provisioned_size] ||= 8 * Fog::Compute::Ovirt::DISK_SIZE_TO_GB
            options[:format] ||= OvirtSDK4::DiskFormat::COW
            options[:sparse] = true unless options[:sparse].present
```
becomes:
```ruby
            options[:disk][:provisioned_size] ||= options.delete(:provisioned_size, 8 * Fog::Compute::Ovirt::DISK_SIZE_TO_GB)
            options[:disk][:format] ||= options.delete(:format, OvirtSDK4::DiskFormat::COW)
            options[:disk][:sparse] ||= options.delete(:sparse, true)
```

PR #53, opened 2018-12-27
Refs #24808 - Fix boolean values in v4

shiramax:


lib/fog/ovirt/requests/compute/v4/add_volume.rb
```diff
@@ -16,25 +16,23 @@ def add_volume(id, options = {})
...
           def add_options_defaults(options)
             options = options.dup
+            options = convert_string_to_bool(options)
             search = options[:search] || format("datacenter=%<datacenter>s", :datacenter => datacenter)
-            options[:bootable] = options.delete(:bootable) == "true"
+            options[:bootable] = options.delete(:bootable)
             options[:interface] ||= OvirtSDK4::DiskInterface::VIRTIO
             options[:provisioned_size] = options[:size_gb].to_i * Fog::Compute::Ovirt::DISK_SIZE_TO_GB if options[:size_gb]
-
+            options[:sparse] ||= true if options[:sparse].nil?
```
orrabin: no need for the if when you are using `||`

lib/fog/ovirt/compute/v4.rb
```diff
@@ -107,6 +107,17 @@ def get_attr_value(value, opts)
             end
           end
           # rubocop:enable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
+
+          def convert_string_to_bool(opts)
+            opts.each do |key, value|
```
orrabin: maybe searching for the values that are "true", "false" instead of the `each`?

lib/fog/ovirt/requests/compute/v4/add_volume.rb
```diff
@@ -16,25 +16,23 @@ def add_volume(id, options = {})
           # rubocop:disable Metrics/AbcSize
           def add_options_defaults(options)
             options = options.dup
+            options = convert_string_to_bool(options)
             search = options[:search] || format("datacenter=%<datacenter>s", :datacenter => datacenter)
-            options[:bootable] = options.delete(:bootable) == "true"
+            options[:bootable] = options.delete(:bootable)
```
orrabin, line 21: what about if this is nil? the previous behavior would return `false` but the `convert_string_to_bool` method wouldn't add it so now we would be passing nil.

orrabin (changes requested):

ShimShtein (changes requested):
I think we are missing `vm_ticket` too. It has a boolean parameter.

Besides that :+1:

lib/fog/ovirt/requests/compute/v4/add_volume.rb
shiramax: As we discussed, The bootable flag doesn't have a default value. if he will receive nill it's the same as false.

shiramax:
@orrabin please review again :)

ShimShtein (approved):

orrabin (approved):

orrabin:
@shiramax rubocop is unhappy, could you please take a look?
once tests are green this is ready to merge.

orrabin:
Thanks @shiramax !

PR #61, opened 2019-03-09
Adding get_volume request

orrabin:


ShimShtein:
@orrabin, It seems that volumes are first rate citizen in fog-ovirt. I think they deserve [collection](https://fog.io/about/structure.html#collections) support. Even if it will take too much refactoring (I don't think so, since we can add all those services alongside to existing ones and gradually deprecate the older interface), we still can use [models](https://fog.io/about/structure.html#models) and not pure attribute hashes as return values.

orrabin:
@ShimShtein as discussed - it does have a collection support that returns `get_volume` but no one implemented it.
For returning the models in general I agree we should refactor this but for all the models.

shiramax:
@orrabin can you explain the initiative for this PR? are we lucking some functionality?

orrabin:
@shiramax I wanted to simplify this: https://github.com/theforeman/foreman/blob/develop/app/models/compute_resources/foreman/model/ovirt.rb#L312 by getting the volume by id. We have the option in the sdk but we didn't implement it here.

ShimShtein:
> I wanted to simplify this: https://github.com/theforeman/foreman/blob/develop/app/models/compute_resources/foreman/model/ovirt.rb#L312 by getting the volume by id. We have the option in the sdk but we didn't implement it here.

Actually it would be better to make a single request to get all volumes and then match those one by one (as the code in Foreman does right now). Using the method proposed in this PR will result in N+1 queries to oVirt (and those are quite expensive).

@orrabin, I don't mind getting this in, but specifically for the purpose you have stated, it's not a good fit.

shiramax:
@ShimShtein @orrabin what's the status of this PR?

orrabin:
@ShimShtein we have a request for `list_volumes` which could be used but if you want a single one you can use the request added by this pr.

ShimShtein:
ACK from me, looks good.

@orrabin, for future references to code, could you please use absolute links (the ones with hash id in it). Now the code has changed, so it's impossible to see what were you linking.

shiramax (approved):

Issue #62, opened 2019-03-17
Undefined method `each` in convert_string_to_bool


ShimShtein:
While creating a host from template with Foreman I am getting:

```
2019-03-17T17:15:49 [W|app|1ff89b4b] »Failed to create a compute oris ovirt4 (oVirt) instance shim-eldon-ganong.dummy: undefined method `each' for nil:NilClass
 |  «
 | NoMethodError: undefined method `each' for nil:NilClass
 | /usr/local/rvm/gems/ruby-2.4.1@foreman/gems/fog-ovirt-1.1.6/lib/fog/ovirt/compute/v4.rb:114:in `convert_string_to_bool'
 | /usr/local/rvm/gems/ruby-2.4.1@foreman/gems/fog-ovirt-1.1.6/lib/fog/ovirt/compute/v4.rb:122:in `block (2 levels) in convert_string_to_bool'
 | /usr/local/rvm/gems/ruby-2.4.1@foreman/gems/fog-ovirt-1.1.6/lib/fog/ovirt/compute/v4.rb:122:in `map'
 | /usr/local/rvm/gems/ruby-2.4.1@foreman/gems/fog-ovirt-1.1.6/lib/fog/ovirt/compute/v4.rb:122:in `block in convert_string_to_bool'
 | /usr/local/rvm/gems/ruby-2.4.1@foreman/gems/fog-ovirt-1.1.6/lib/fog/ovirt/compute/v4.rb:114:in `each'
 | /usr/local/rvm/gems/ruby-2.4.1@foreman/gems/fog-ovirt-1.1.6/lib/fog/ovirt/compute/v4.rb:114:in `convert_string_to_bool'
 | /usr/local/rvm/gems/ruby-2.4.1@foreman/gems/fog-ovirt-1.1.6/lib/fog/ovirt/requests/compute/v4/create_vm.rb:27:in `create_vm'
 | /usr/local/rvm/gems/ruby-2.4.1@foreman/gems/fog-ovirt-1.1.6/lib/fog/ovirt/models/compute/server.rb:179:in `save'
 | /usr/local/rvm/gems/ruby-2.4.1@foreman/gems/fog-core-1.45.0/lib/fog/core/collection.rb:50:in `create'
 | /home/shim/Documents/foreman/foreman/app/models/compute_resource.rb:195:in `create_vm'
 | /home/shim/Documents/foreman/foreman/app/models/compute_resources/foreman/model/ovirt.rb:282:in `create_vm'
```

[...]

PR #63, opened 2019-03-25
Converts string to bool only if the string exists

orrabin:
This fixes the issue: https://github.com/fog/fog-ovirt/issues/62


shiramax (approved):
@orrabin, can you please add a test for that?

orrabin:
@shiramax not really with our current test structure, I'd have to add a test through foreman for this that sends nil to `convert_string_to_bool` and doesn't raise an exception.

shiramax (approved):

Issue #64, opened 2019-04-02
Error When i try to run the app  syntax error, unexpected ',' (SyntaxError) ...e_search_by_datacenter(search:, datacenter:, page: nil)


mariohmol:
Hi.

i have an up where runs the bundle install , but when i run via `rails s` i get this error

Any ideas?

```
MacBook-Pro-de-Gil:app user$ rails s
/Users/user/.rvm/gems/ruby-2.0.0-p451/gems/activesupport-4.0.1/lib/active_support/dependencies.rb:229:in `require': /Users/user/.rvm/gems/ruby-2.0.0-p451/gems/fog-ovirt-1.1.6/lib/fog/ovirt/compute/v4.rb:193: syntax error, unexpected ',' (SyntaxError)
...e_search_by_datacenter(search:, datacenter:, page: nil)
...                               ^
/Users/user/.rvm/gems/ruby-2.0.0-p451/gems/fog-ovirt-1.1.6/lib/fog/ovirt/compute/v4.rb:207: syntax error, unexpected keyword_end, expecting end-of-input
	from /Users/user/.rvm/gems/ruby-2.0.0-p451/gems/activesupport-4.0.1/lib/active_support/dependencies.rb:229:in `block in require'
	from /Users/user/.rvm/gems/ruby-2.0.0-p451/gems/activesupport-4.0.1/lib/active_support/dependencies.rb:214:in `load_dependency'
	from /Users/user/.rvm/gems/ruby-2.0.0-p451/gems/activesupport-4.0.1/lib/active_support/dependencies.rb:229:in `require'
	from /Users/user/.rvm/gems/ruby-2.0.0-p451/gems/fog-ovirt-1.1.6/lib/fog/ovirt/compute.rb:54:in `<class:Ovirt>'
	from /Users/user/.rvm/gems/ruby-2.0.0-p451/gems/fog-ovirt-1.1.6/lib/fog/ovirt/compute.rb:3:in `<module:Compute>'
	from /Users/user/.rvm/gems/ruby-2.0.0-p451/gems/fog-ovirt-1.1.6/lib/fog/ovirt/compute.rb:2:in `<module:Fog>'
	from /Users/user/.rvm/gems/ruby-2.0.0-p451/gems/fog-ovirt-1.1.6/lib/fog/ovirt/compute.rb:1:in `<top (required)>'
[...]

PR #65, opened 2019-04-04
New function giving syntax error in 2.0x ruby

mariohmol:
I think this will fix this #64 

thanks

### DIFF
--- a/lib/fog/ovirt/compute/v4.rb
+++ b/lib/fog/ovirt/compute/v4.rb
@@ -191,7 +191,7 @@ module Fog
           end
 
           def create_search_by_datacenter(search: nil, datacenter: nil, page: nil)
-            unless datacenter raise ::Fog::Ovirt::Errors::OvirtEngineError('Cannot create search request by datacenter, when datacenter is nil`)
+            unless datacenter raise ::Fog::Ovirt::Errors::OvirtEngineError('Cannot create search request by datacenter, when datacenter is nil')
             search ||= ""
             search += " datacenter=#{datacenter}" if datacenter
             search += " page #{page}" if page
--- a/lib/fog/ovirt/compute/v4.rb
+++ b/lib/fog/ovirt/compute/v4.rb
@@ -190,7 +190,7 @@ module Fog
             @blank_template ||= client.system_service.get.special_objects.blank_template
           end
 
-          def create_search_by_datacenter(search:, datacenter:, page: nil)
+          def create_search_by_datacenter(search: nil, datacenter: nil, page: nil)
             search ||= ""
             search += " datacenter=#{datacenter}" if datacenter
             search += " page #{page}" if page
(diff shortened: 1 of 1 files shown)


shiramax (approved):
Seems good to me, 
@orrabin, @ShimShtein  One of you can review as well?

lib/fog/ovirt/compute/v4.rb
```diff
@@ -190,7 +190,7 @@ def blank_template
             @blank_template ||= client.system_service.get.special_objects.blank_template
           end
 
-          def create_search_by_datacenter(search:, datacenter:, page: nil)
+          def create_search_by_datacenter(search: nil, datacenter: nil, page: nil)
```
ShimShtein, line 193: Can you please add guard conditions to the function? You are relaxing requirements for the function, but it should not be possible to call it without these two parameters.

ShimShtein (changes requested):
Indeed, required kwargs were introduced in `ruby 2.1`, so it's an invalid syntax for `2.0`.

lib/fog/ovirt/compute/v4.rb
mariohmol: any ideas on how to make it?

ShimShtein: Hey, sorry for the long wait, this one slipped under my radar.

It looks like the only parameter that is mandatory is `datacenter`, so I would add an if statement:

``` ruby
def create_search_by_datacenter(search: nil, datacenter: nil, page: nil)
  unless datacenter raise ::Fog::Ovirt::Errors::OvirtEngineError('Cannot create search request by datacenter, when datacenter is nil`)
```

or something similar.

mariohmol: Allright!

Updated in this PR

thanks

lib/fog/ovirt/compute/v4.rb
```diff
@@ -190,7 +190,8 @@ def blank_template
...
+            unless datacenter raise ::Fog::Ovirt::Errors::OvirtEngineError('Cannot create search request by datacenter, when datacenter is nil`)
```
shiramax:
```suggestion
            unless datacenter raise ::Fog::Ovirt::Errors::OvirtEngineError('Cannot create search request by datacenter, when datacenter is nil')
```

lib/fog/ovirt/compute/v4.rb
```diff
@@ -190,7 +190,8 @@ def blank_template
...
             search ||= ""
             search += " datacenter=#{datacenter}" if datacenter
```
shiramax, line 196:
```suggestion
            search += " datacenter=#{datacenter}" 
```

shiramax (changes requested):

mariohmol:
Sorry, didnt see this request, for me its ok!

PR #68, opened 2019-05-15
Support storage domain lease and pass discard

zozoh94:
Support storage domain lease with HA VM and disk attachment pass discard

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -9,7 +9,7 @@
 # Offense count: 1
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 150
+  Max: 180
 
 # Offense count: 2
 # Configuration parameters: CountComments.
--- a/lib/fog/ovirt/models/compute/server.rb
+++ b/lib/fog/ovirt/models/compute/server.rb
@@ -32,6 +32,7 @@ module Fog
         attribute :ips
         attribute :ha
         attribute :ha_priority
+        attribute :lease_storage_domain
         attribute :clone
         attribute :disks
 
--- a/lib/fog/ovirt/models/compute/volume.rb
+++ b/lib/fog/ovirt/models/compute/volume.rb
@@ -17,6 +17,7 @@ module Fog
         attribute :quota
         attribute :alias
         attribute :wipe_after_delete
+        attribute :attachment_pass_discard
 
         def size_gb
           attributes[:size_gb] ||= attributes[:size].to_i / Fog::Compute::Ovirt::DISK_SIZE_TO_GB if attributes[:size]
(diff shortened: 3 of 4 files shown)


zozoh94:
Hi ! Someone can explain me why Travis is failing ?

shiramax:
@zozoh94 thanks for you PR!
this PR: https://github.com/fog/fog-ovirt/pull/69 should fix the rubocop errors.

shiramax (changes requested):
@zozoh94 
I think it's better to wait until #69 will merge and then you will be able to rebase your PR.
Also, do you have a PR for foreman too?

zozoh94:
@shiramax  As you want. I'll do the PR for Foreman when this one will be merged.

shiramax:
@zozoh94 
We have a strict policy for adding oVirt features to Foreman, 
Because we can't have all oVirt features in Foreman,
Can you explain why do you need to support the leases? Why using templates it's not enough?

JBaptisme:
Hello, I will try to explain benefit to add lease choice:
Lease is directly linked to High available. 
I think, a lease should be placed into first (bootable one for exemple) disk, if High available is checked.
I understand to use oVirt Template, but in this case, HA option should not be present in Foreman.

Moreover, Foreman provide good granularity regarding Storage Domain choice when provision VM, and not related to oVirt existing templates.
Lease is same think. It should be a list choice as Storage Domain. 
If chosen, Lease should be same as one of Storage Domain disk (preferably booting one)

Thanks for your time,

Regards

shiramax:
@zozoh94 @JBaptisme 
 thanks for your comments, 
I still not sure I completely understand, 
can you create a template in ovirt that includes the properties you want? including the lease the ha part? 
and then create from that image a virtual machine in foreman?

JBaptisme:
Hello @shiramax ,

Yes, It's possible to create a template into oVirt. 
But with ovirt template , it mean "Target Storage Domain for VM Lease" will be static ( it will be one of Storage domain, statically defined and that's all)...

If Target Storage Domain for VM Lease is available into Foreman, (like in oVirt template), it will be possible to choose destination storage domain for VM Lease for each Foreman VM created unlike template way.
For option Target Storage Domain for VM Lease , menu should me the same as Disk Storage domain

I hope It will be clearer ,

shiramax:
thanks @zozoh94 @JBaptisme  :)
sounds ok to me, 
can you please explain about the discard? and also could you please add the foreman PR so I could test both together?

erezzarum:
For some odd reason the lease does not inherit when using a template (at least a full clone one), just ran across this issue (oVirt 4.4) so it's either oVirt which should handle it or Foreman (uses fog)

PR #73, opened 2019-06-24
Add vm resources max values

bagasse:
Related to https://projects.theforeman.org/issues/27086, https://github.com/theforeman/foreman/pull/6852#issuecomment-504726272

Add support of some VM resources max values (vcpus, socket, mem).

I've only oVirt 4.3 instances to test so I don't have any values for previous oVirt version, that's why there is only values for 4.1, 4.2 and 4.3 versions.

### DIFF
--- a/lib/fog/ovirt/compute/base.rb
+++ b/lib/fog/ovirt/compute/base.rb
@@ -1,0 +1,103 @@
+module Fog
+  module Ovirt
+    class Compute
+      class Base
+        # Max resources values for VMs depending on oVirt's version
+        # These values can be found by running the engine-config command on the engine
+        # with following config keys:
+        # - MaxNumOfVmCpus
+        # - MaxNumOfVmSockets
+        # - VM64BitMaxMemorySizeInMB
+        # - VM32BitMaxMemorySizeInMB,
+        #
+        # Eg:
+        #
+        # engine-config --get MaxNumOfVmCpus
+        # MaxNumOfVmCpus: 288 version: 4.1
+        # MaxNumOfVmCpus: 384 version: 4.2
+        # MaxNumOfVmCpus: 384 version: 4.3
+        #
+        VM_MAX_VALUES = {
+          # Default values for unknown version
+          "unknown" => { :max_num_of_vm_cpus => 240,
+                         :max_num_of_vm_sockets => 16,
+                         :vm64_bit_max_memory_size_in_mb => 4_194_304,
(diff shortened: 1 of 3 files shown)


lib/fog/ovirt/models/compute/cluster.rb
```diff
@@ -5,8 +5,59 @@ class Cluster < Fog::Model
...
+        # engine-config --get MaxNumOfVmCpus
+        # MaxNumOfVmCpus: 288 version: 4.1
+        # MaxNumOfVmCpus: 384 version: 4.2
+        # MaxNumOfVmCpus: 384 version: 4.3
+        #
+        VM_MAX_VALUES = {
+          "4.1" => { :max_num_of_vm_cpus => 288,
+                     :max_num_of_vm_sockets => 16,
+                     :vm64_bit_max_memory_size_in_mb => 4_194_304,
+                     :vm32_bit_max_memory_size_in_mb => 2048 },
```
ezr-ondrej: Shouldn't this (and all three others) be 20_480?

ezr-ondrej (commented):
LGTM :+1: , just a typo :)

lib/fog/ovirt/models/compute/cluster.rb
bagasse: You are right, sorry.

bagasse:
@ezr-ondrej thanks for the review, that's fixed :)

bagasse:
@ezr-ondrej, sorry, totally reworked this to attach it to Datacenter instead of cluster as @shiramax suggested in https://github.com/theforeman/foreman/pull/6852

orrabin:
@bagasse have you tried removing the max value all together?
oVirt should send a message when we pass the max and unless there is a way to get this dynamically from the API (which I understand from the comment linked in the description there isn't) I'm concerned about the values changing each version and how to keep up with the changes.

bagasse:
@orabin, thanks for suggestion, on the foreman side, as I saw, this is only used on the compute vm attributes form to provide a user feedback before submitting the host form, so yes maybe the solution is to remove this entirely from the foreman form side and that's it. Both solutions are fine to me, one need more code on fog-ovirt side and so more maintenance, the other remove some code on foreman side so decrease maintenance but remove some user feedback on vm creation.

orrabin:
@bagasse I agree it removes some feedback but in this case it might be worth it to avoid the cost of possibly breaking for each new version on oVirt.

bagasse:
@orrabin, ok thanks for feedback, closing it

PR #90, opened 2020-03-08
Disable Metrics/AbcSize cop

shiramax:
Since we are already disabling it in each function specifically.


orrabin (approved):

orrabin:
Thanks @shiramax !

PR #93, opened 2020-07-10
Set minimum version of ovirt-engine-sdk-ruby

gekorob:
Since the version 4.3.1 of the sdk there's a new option in the
VmService#start method (:use_initialization) that could help to create
a solution for the problem reported here

https://community.theforeman.org/t/foreman-windows-and-ovirt/19365


shiramax (approved):
thanks @gekorob !

PR #94, opened 2020-07-10
use_initialization in vm start 

gekorob:
Since the version 4.3.1 of the ovirt-engine-sdk-ruby is possible to start a vm passing the use_initialization param.
This can be an OS agnostic way to pass init data to the starting vm

depends on:
https://github.com/fog/fog-ovirt/pull/93

### DIFF
--- a/lib/fog/ovirt/compute/v4.rb
+++ b/lib/fog/ovirt/compute/v4.rb
@@ -9,6 +9,7 @@ module Fog
 
         request :vm_action
         request :vm_start_with_cloudinit
+        request :vm_start_with_initialization
         request :destroy_vm
         request :create_vm
         request :update_vm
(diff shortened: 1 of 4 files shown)


shiramax:
@gekorob thanks for your contribution!
1. could you please rebase?
2. could you explain a bit about use_initialization and why you need it? 
3. Do you have another PR in foreman core?

gekorob:
@shiramax 
1. done
2. it was in response to the problem reported in https://community.theforeman.org/t/foreman-windows-and-ovirt/19365
3. no, I was working to manageiq-providers-ovirt and ovirt-engine-sdk-ruby in a very recent past.

Rubocop is failing for too many lines in the class, but I think it's a constraint difficult to satisfy in this case.
If you think this PR is not good to merge feel free to close it.

shiramax:
@gekorob I don't think that PR is not good for merging, but as of now no one is using this function start_with_initialization, so I wonder if another PR in foreman core is needed.

PR #95, opened 2020-09-08
Fixes #30793 - Remove API v3

shiramax:


lib/fog/ovirt/compute.rb
```diff
@@ -50,73 +50,53 @@ def respond_to_missing?(method_name, include_private = false)
...
-          @client = Fog::Ovirt::Compute::V3.new(options)
-        end
-        # rubocop:enable Style/ConditionalAssignment
+        @client = Fog::Ovirt::Compute::V4.new(options)
       end
 
-      # rubocop:disable Style/MethodMissingSuper, Style/MissingRespondToMissing
       def method_missing(symbol, *args)
         @client.__send__(symbol, *args)
+        super
```
ShimShtein: Why do you want to call `Object.method_missing` after calling the same method on `@client`?

lib/fog/ovirt/compute.rb
```diff
@@ -50,73 +50,53 @@ def respond_to_missing?(method_name, include_private = false)
...
       end
 
       def respond_to?(symbol, include_all = false)
         @client.respond_to?(symbol, include_all)
       end
-      # rubocop:enable Style/MethodMissingSuper, Style/MissingRespondToMissing
+      # rubocop:enable
```
ShimShtein: You can remove re-enabling of rubocop here.

lib/fog/ovirt/compute.rb
```diff
@@ -50,73 +50,53 @@ def respond_to_missing?(method_name, include_private = false)
...
-            @client = Fog::Ovirt::Compute::V3::Mock.new(options)
-          end
+          Fog::Ovirt::Compute::V4::Mock.include Fog::Ovirt::Compute::Collections
+          @client = Fog::Ovirt::Compute::V4::Mock.new(options)
         end
 
-        # rubocop:disable Style/MethodMissingSuper, Style/MissingRespondToMissing
         def method_missing(symbol, *args)
           @client.__send__(symbol, *args)
+          super
```
ShimShtein: Why `super` here?

lib/fog/ovirt/compute.rb
```diff
@@ -50,73 +50,53 @@ def respond_to_missing?(method_name, include_private = false)
...
-            @client = Fog::Ovirt::Compute::V3::Real.new(options)
-          end
+          Fog::Ovirt::Compute::V4::Real.include Fog::Ovirt::Compute::Collections
+          @client = Fog::Ovirt::Compute::V4::Real.new(options)
         end
 
-        # rubocop:disable Style/MethodMissingSuper, Style/MissingRespondToMissing
         def method_missing(symbol, *args)
           @client.send(symbol, *args)
+          super
```
ShimShtein: Why `super` here?

lib/fog/ovirt/compute.rb
shiramax: without the super, I'm keep getting rubocop errors: 
lib/fog/ovirt/compute.rb:60:7: C: Style/MethodMissingSuper: When using method_missing, fall back on super.
      def method_missing(symbol, *args) ...
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/fog/ovirt/compute.rb:90:9: C: Style/MethodMissingSuper: When using method_missing, fall back on super.
        def method_missing(symbol, *args)

lib/fog/ovirt/compute.rb
shiramax: same.

lib/fog/ovirt/compute/v4.rb
```diff
@@ -1,3 +1,5 @@
+require "active_support"
```
ShimShtein, line 1: `active_support` is part of rails, do you have to use it here? I think Fog does not require it.

lib/fog/ovirt/compute.rb
```diff
@@ -50,73 +50,54 @@ def respond_to_missing?(method_name, include_private = false)
...
-          @client = Fog::Ovirt::Compute::V3.new(options)
-        end
-        # rubocop:enable Style/ConditionalAssignment
+        @client = Fog::Ovirt::Compute::V4.new(options)
       end
 
-      # rubocop:disable Style/MethodMissingSuper, Style/MissingRespondToMissing
       def method_missing(symbol, *args)
         @client.__send__(symbol, *args)
+        super
```
ShimShtein: Leftover?

Gemfile
```diff
@@ -1,3 +1,3 @@
 source "https://rubygems.org"
-
+gem "activesupport"
```
ShimShtein: `activesupport` is part of rails. Do you really need it? If it's needed, I suggest moving the requirement to `.gemspec`.

ShimShtein (changes requested):
I suppose you didn't want to add `Gemfile.local.rb`. You should modify `.gitignore` to exclude it from commits.

Gemfile
shiramax: I need it for running the tests, I will move it to the .gemspec. thanks.

lib/fog/ovirt/compute.rb
shiramax: ye, sorry. I will change it.

shiramax:
@ShimShtein thanks for the review, I have addressed all your comments.

lib/fog/ovirt/compute/v4.rb
shiramax: If I remove the active_support,  when running the tests, I'm failing for: 
https://github.com/fog/fog-ovirt/blob/master/lib/fog/ovirt/compute/v4.rb#L55 


undefined method `present?' for #<OvirtSDK4::VnicProfile:0x00007f1750edc740> (NoMethodError)

fog-ovirt.gemspec
```diff
@@ -16,10 +16,11 @@ Gem::Specification.new do |spec|
   spec.license       = "MIT"
 
   spec.files         = `git ls-files -z`.split("\x0")
-  spec.test_files    = spec.files.grep(%r{^tests\/})
+  spec.test_files    = spec.files.grep(%r{^tests/})
 
   spec.require_paths = ["lib"]
-
+  spec.required_ruby_version = ">= 2.0.0"
+  spec.add_dependency("activesupport")
```
ShimShtein, line 23: If you need it only for running tests, you can have it as development dependency:
```suggestion
  spec.add_development_dependency("activesupport")
```

lib/fog/ovirt/compute/v4.rb
ShimShtein: I still don't see a reason to require `active_support` here.
If you need it for testing, it would be better to require it in test helpers. Although if the code should run without `active_support`, we should run the tests without it too. Otherwise we can slip some `active_suport` stuff in our code, and it won't be detected until the code is executed in prod setup.

ShimShtein (changes requested):

shiramax:
@ShimShtein we talked offline and we decided to remove active_support in a different PR. It would be great if you can re-review this PR. thanks!

lib/fog/ovirt/compute/v4.rb
```diff
@@ -53,7 +56,6 @@ def shared_ovirt_attrs(obj)
 
               if key == :network
                 opts[key] = obj.vnic_profile.present? ? yield(obj) : value
-                next
```
ShimShtein: this is a change to behaviour, is it intentional?

ShimShtein (approved):
Not tested, but ACK from me

yifatmakias (approved):
Works as expected and looks good :+1:

orrabin:
This received 2 acks so I'm merging.
Thanks @shiramax !

Also thank you to @ShimShtein and @yifatmakias for reviewing!

PR #96, opened 2020-09-17
Fixes #30835 - Allow to filter on multi SD types

shiramax:


lib/fog/ovirt/requests/compute/v4/storage_domains.rb
```diff
@@ -9,7 +9,7 @@ def storage_domains(filter = {})
             filter[:search] = create_search_by_datacenter(:search => filter[:search], :datacenter => datacenter_hash[:name])
             client.system_service.storage_domains_service.list(filter).collect do |sd|
               # Filter by role is not supported by the search language. The work around is to list all, then filter.
-              role_filter.nil? || sd.type == role_filter ? sd : nil
+              role_filter.nil? || (role_filter.include? sd.type) ? sd : nil
```
tbrisker: You can wrap the `role_filter` to preserve backward-compatibility. also I would suggest different way of parenthesis to make the trenary condition clearer:
```suggestion
              (role_filter.nil? || Array(role_filter).include?(sd.type)) ? sd : nil
```
cc: @orrabin

orrabin (approved):

orrabin:
Thanks @shiramax !

PR #104, opened 2022-02-09
Fixes # 34422  - Fix sockets attributes

adiabramovitch:


lib/fog/ovirt/requests/compute/v4/update_vm.rb
```diff
@@ -9,7 +9,7 @@ def update_vm(attrs)
             vm_service = client.system_service.vms_service.vm_service(attrs[:id])
 
             if attrs[:cores].present?
-              cpu_topology = OvirtSDK4::CpuTopology.new(:cores => attrs[:cores], :sockets => "1")
+              cpu_topology = OvirtSDK4::CpuTopology.new(:cores => attrs[:cores], :sockets => attrs[:sockets] || 1)
```
ShimShtein, line 12: The type of the default value is changing from string to int, is it intentional?

ShimShtein (changes requested):

ShimShtein (approved):
Since we have tried it, and it worked as expected, I think we're good.

ShimShtein:
Merged, thanks @adiabramovitch !

PR #107, opened 2023-12-08
Fix storage domain issue when attaching LUN

sbernhard:
Thanks @Manisha15

### DIFF
--- a/lib/fog/ovirt/requests/compute/v4/list_vm_volumes.rb
+++ b/lib/fog/ovirt/requests/compute/v4/list_vm_volumes.rb
@@ -13,7 +13,12 @@ module Fog
               interface = attachment.interface
               attachment_disk.bootable = bootable if attachment_disk.bootable.nil?
               attachment_disk.interface = interface if attachment_disk.interface.nil?
-              attachment_disk.storage_domain = attachment_disk.storage_domains[0].id
+              attachment_disk.storage_domain = begin
+                attachment_disk.storage_domains.first&.id
+              rescue StandardError => e
+                Fog::Logger.warning("Error parsing attachment_disk.storage_domains - #{e}. Fallback to 0")
+                0
+              end
               ovirt_attrs attachment_disk
             end
           end


sbernhard:
Workaround for https://projects.theforeman.org/issues/30199

Review would be nice @orrabin  / @ShimShtein

sbernhard:
Thoughts on this @ShimShtein ?

lib/fog/ovirt/requests/compute/v4/list_vm_volumes.rb
```diff
@@ -13,7 +13,12 @@ def list_vm_volumes(vm_id)
               interface = attachment.interface
               attachment_disk.bootable = bootable if attachment_disk.bootable.nil?
               attachment_disk.interface = interface if attachment_disk.interface.nil?
-              attachment_disk.storage_domain = attachment_disk.storage_domains[0].id
+              attachment_disk.storage_domain = begin
+                                                 attachment_disk.storage_domains[0].id || 0
```
ShimShtein: This assumes there is always a storage domain with id=0.
Is it the correct behavior? I think I would go with `nil`

ShimShtein: Also we don't have to add a `rescue` statement according to the error described in the issue:
```ruby
attachment_disk.storage_domains&.first&.id
```
This will do the trick of being defensive here.

sbernhard: Actually, we just "implemented" what was said is a "good" workaround: https://bugzilla.redhat.com/show_bug.cgi?id=1882404#c9
(better than to have it completely broken)

ShimShtein: Can you try and see if a `nil` value will also work as a workaround? I think putting magic values is prone to errors later on.

sbernhard:
ping @ShimShtein :-)

lib/fog/ovirt/requests/compute/v4/list_vm_volumes.rb
```diff
@@ -13,7 +13,12 @@ def list_vm_volumes(vm_id)
               interface = attachment.interface
               attachment_disk.bootable = bootable if attachment_disk.bootable.nil?
               attachment_disk.interface = interface if attachment_disk.interface.nil?
-              attachment_disk.storage_domain = attachment_disk.storage_domains[0].id
+              attachment_disk.storage_domain = begin
+                                                 attachment_disk.storage_domains[0].id || nil
```
ShimShtein:
```suggestion
                                                 attachment_disk.storage_domains&.first&.id || nil
```

ShimShtein (changes requested):
I think we don't want to push `0` as the storage domain id, if `storage_domains` returned nil or and empty list. If we don't use the safe navigation operator here, we will end up in the `rescue` block (which is not desired, I think)

I can understand that in the case of an actual error, we can assume there is at least one domain, but we can't get the list, so we assume there is only one.

sbernhard:
> I think we don't want to push `0` as the storage domain id, if `storage_domains` returned nil or and empty list. If we don't use the safe navigation operator here, we will end up in the `rescue` block (which is not desired, I think)
> 
> I can understand that in the case of an actual error, we can assume there is at least one domain, but we can't get the list, so we assume there is only one.

This would then not work. The original issue was:
image

So, `attachment_disk.storage_domains` is nil and the access to "[0]" crashes. In this case, and only in this case it should return with "0" which would then work in case of direct attached LUN disks.

Its still better to return 0 instead of crashing.

ShimShtein (approved):
Didn't test it, but now the code looks good.

I see three cases here:
1. `storage_domains` returns one or more domains (the "happy" path), then the `storage_domain` property would be set to the first domain from the list
2. `storage_domains` returns an empty list, then the `storage_domain` property should be set to `nil` (we didn't find a good candidate).
3. `storage_domains` throws an exception, then we want to guess the storage domain id. We are setting the `storage_domain` property to `0`, since it's our best guess.

dosas:
@sbernhard There were quite some changes on the default branch. Could you rebase to enable the CI?

sbernhard:
@ShimShtein I had a look at the code again.
1. in case of there is 'attachment_disk_storage_domains' first entry with an id  => use this id 
2. in case of there is 'attachment_disk_storage_domains' first entry but no id field is found => nil should be returned
3. in case of NO  'attachment_disk_storage_domains' first entry, fallback to 0


This is what https://bugzilla.redhat.com/show_bug.cgi?id=1882404#c9 suggestes and works for us.

dosas (approved):

sbernhard:
@ShimShtein can you have a look at this again, please?

dosas:
@ShimShtein Do you think you could merge this? We do not have the necessary permissions.

dosas (approved):

PR #108, opened 2024-04-30
Fix open struct error

dosas:
uninitialized constant Fog::Ovirt::Compute::V4::Shared::OpenStruct (NameError)

Fog::Ovirt::Compute.new | template model (ovirt) test is failing because of this

### DIFF
--- a/lib/fog/ovirt/compute/v4.rb
+++ b/lib/fog/ovirt/compute/v4.rb
@@ -1,5 +1,6 @@
 require "active_support"
 require "active_support/core_ext"
+require "ostruct"
 
 module Fog
   module Ovirt


dosas:
@ShimShtein Would you have time to review this and my other PRs I made?

lib/fog/ovirt/compute/v4.rb
```diff
@@ -88,6 +88,7 @@ def get_attr_value(value, opts)
               opts[:ha] = value.enabled
               value
             when OvirtSDK4::TemplateVersion
+              require "ostruct"
```
ShimShtein: I think we can move it to the head of the file.
We are not loading too much here that it's worth hiding this `require` in the code.

dosas: I was trying to follow the existing coding style. This pattern is used throughout the file.

ShimShtein: It's a bit different case: there are `require "ovirtsdk4"` statements in code, because we had support for both v3 and v4, and you had to select which one are you using.
In the current case, it's something that is related to the data structure that is returned, so it is more similar to the `require "active_support"` statement that we have in the header

dosas: Thanks for clarifying this. Updated the PR. 

Would really appreciate it if you could have a look at my other PRs as well.

ShimShtein (approved):
LGTM.

lib/fog/ovirt/compute/v4.rb
ShimShtein: Just tag me on the PR, and hopefully I will get to it

dosas:
@ShimShtein I do not have permission to merge this PR, can you do it?

PR #109, opened 2024-04-30
Exclude vendor dirs in .rubocop.yaml

dosas:


dosas:

ShimShtein (approved):

PR #110, opened 2024-04-30
Fix rubocop yaml config error

dosas:
.rubocop.yml: Metrics/LineLength has the wrong namespace - should be Layout


dosas:

ShimShtein (approved):

PR #111, opened 2024-05-02
Explicitly list files and dirs included in gemspec (test_)files

dosas:
With the catch all approach there were a lot of files included that do not need to be in the gem.
I.e. Jenkinsfile, all dot files and dirs related to CI configuration, and rubocop linting ...

This causes problems when i.e. building rpm packages where all of these files need to be explicitly excluded


dosas:

ShimShtein (approved):

PR #114, opened 2024-08-21
Update rubocop and enable it in CI

dosas:


dosas:
@ShimShtein This also enables the rubocop in the CI

.github/workflows/ci.yml
```diff
@@ -32,3 +31,23 @@ jobs:
         bundler-cache: true # runs 'bundle install' and caches installed gems automatically
     - name: Run tests
       run: bundle exec rake
+
+  lint:
```
ShimShtein: For performance sake, can we add rubocop just as a step before the tests? It would spare a full spin up of a machine just for running rubocop.

fog-ovirt.gemspec
```diff
@@ -30,6 +29,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "bundler"
   spec.add_development_dependency "pry"
   spec.add_development_dependency "rake"
-  spec.add_development_dependency "rubocop", "~> 0.52"
+  spec.add_development_dependency "rubocop"
```
ShimShtein: Let's freeze rubocop version. It tends to change under our feet and fail CI.
If you think we would like to keep rubocop up to date, I would recommend adding a separate CI job that will test against the latest rubocop once in a while and we will monitor it.

dosas: Fixed the version, we could use dependabot for this?

.github/workflows/ci.yml
dosas: Yes that makes sense.

ShimShtein (approved):

fog-ovirt.gemspec
ShimShtein: I don't have experience with dependabot, but I am not against it. We can always disable it later :shrug:

ShimShtein:
@dosas I see rubocop started to fail in CI. Any idea why?

PR #116, opened 2024-10-14
v2.0.3

geemus:
Dropped changelog from gemspec as it doesn't appear to exist and building gem fails when specified files are absent.


geemus:
Once this is merged I can build/release.

dosas (approved):

geemus:
Since this was blocked by a reviewer with commit, I (at least for now) gave myself permission for direct pushes. This is inline with other fog repos, and unblocked. Having done that I directly pushed and released, so this version should now be out and available.

dosas:
> Since this was blocked by a reviewer with commit, I (at least for now) gave myself permission for direct pushes. This is inline with other fog repos, and unblocked. Having done that I directly pushed and released, so this version should now be out and available.

Thank you! https://rubygems.org/gems/fog-ovirt/versions/2.0.3

